PR #1, opened 2023-03-09
CI setup

jamieparkinson:
Runs tests and autoformatting (includes some changes from the latter). The way docker-compose and yarn are set up won't work as-is if we want to do code sharing, but that's OK.


tooling.Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM public.ecr.aws/docker/library/node:18
+
+# Install Terraform (for formatting)
```
jamieparkinson, line 3: Here cos it'll get used eventually

rcantin-w (approved):
Makes sense to me

Issue #2, opened 2023-03-10
content-api: Infrastructure


jamieparkinson:
- [x] An ECS service https://github.com/wellcomecollection/content-api/pull/3
- [x] Behind a load balancer https://github.com/wellcomecollection/content-api/pull/3
- [x] Behind APIGW https://github.com/wellcomecollection/content-api/pull/5
- [x] Behind CloudFront https://github.com/wellcomecollection/platform-infrastructure/pull/372
- [x] CI publishing the image and deploying the service https://github.com/wellcomecollection/content-api/pull/7

pollecuttn:
@jamieparkinson Did you say this one can be closed? I don't have the power to close it.

pollecuttn:
@jamieparkinson Ignore me! Didn't refresh my board before asking.

PR #3, opened 2023-03-10
ECS service behind a load balancer

jamieparkinson:
Part of https://github.com/wellcomecollection/content-api/issues/2

Doesn't include API Gateway or CloudFront or any deployment stuff.

This is largely copied from the catalogue-api repo (although I'm trying to keep the directory structure flatter and simpler, because it's been confusing to find where things live in there)

### DIFF
--- a/infrastructure/ecr.tf
+++ b/infrastructure/ecr.tf
@@ -1,0 +1,7 @@
+resource "aws_ecr_repository" "concepts_api" {
+  name = "weco/content-api"
+
+  lifecycle {
+    prevent_destroy = true
+  }
+}
--- a/infrastructure/ecr.tf
+++ b/infrastructure/ecr.tf
@@ -1,4 +1,4 @@
-resource "aws_ecr_repository" "concepts_api" {
+resource "aws_ecr_repository" "content_api" {
   name = "weco/content-api"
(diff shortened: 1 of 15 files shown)


infrastructure/ecr.tf
```diff
@@ -0,0 +1,7 @@
+resource "aws_ecr_repository" "concepts_api" {
```
alexwlchan, line 1:
```suggestion
resource "aws_ecr_repository" "content_api" {
```

infrastructure/stack/security_groups.tf
```diff
@@ -0,0 +1,37 @@
+data "aws_vpc" "vpc" {
+  id = var.vpc_id
+}
+
+resource "aws_security_group" "vpc_ingress" {
+  name        = "${var.environment}-content-api-vpc-ingress"
+  description = "Allow traffic between services and the rest of the VPC"
```
alexwlchan, line 7: Is this description actually accurate? I know it's copied, but let's not copy round incorrect information.

jamieparkinson: Ha, I had this thought when I copied it - I don't think so either!

PR #6, opened 2023-03-17
Add Elasticsearch config, create data based on Prismic and read from ES on the endpoints

rcantin-w:
`/articles/`
`/articles/Y_i3aBQAACcAwsYW`

The data now gets fetched from Prismic, transformed and sent to ElasticSearch. Our endpoints (above) now get their data from the testing index in ES.

We know this needs to get reorganised within the new `pipeline` folder and will be looking at that next.


api/src/controllers/articles.ts
```diff
@@ -1,73 +1,96 @@
...
+              bool: {
+                should: queryString
+                  ? [
+                      {
+                        match: {
+                          "display.title": {
+                            query: queryString,
+                            analyzer: "whitespace",
+                            operator: "OR",
+                            boost: 1000,
```
jamieparkinson, line 53: As you say, to be left for later, but probably worth removing the unused stuff here
```suggestion
                          "display.title": {
                            query: queryString,
```

api/src/controllers/elastic-search.ts
```diff
@@ -0,0 +1,130 @@
...
+    try {
+      const searchResponse = await clients.prismic.getAllByType<
+        ArticlePrismicDocument & {
+          contentType: ContentType | ContentType[];
+        }
+      >("articles", {
+        // We'll need to consider "webcomics" as well when we do this properly
+        graphQuery: graphQueryArticles,
+        predicates: [
+          prismic.predicate.any("document.type", articlesContentTypes),
```
jamieparkinson, line 99: I can't remember if I asked this before - why do we have a predicate for `document.type` _and_ a `getAllByType`? Isn't the latter doing the type filtering?

api/src/controllers/elastic-search.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  if (!exists) {
+    await elasticClient.indices.create({ index: indexName });
+    console.log(indexName, "was created");
+  } else {
+    console.log("Index", indexName, "already exists");
+  }
+};
+
+const useBulkHelper = async (elasticClient: Client, docs: Article[]) => {
```
jamieparkinson, line 45: `use` is a very vague verb - how about something like `bulkIndexArticles`?

api/src/controllers/elastic-search.ts
```diff
@@ -0,0 +1,130 @@
...
+    return [
+      { index: { _index: indexName, _id: doc.id } },
+      {
+        display: doc as Article,
+        query: { id: doc.id },
+      },
+    ];
+  });
+
+  const bulkResponse = await elasticClient.bulk({ refresh: true, operations });
```
jamieparkinson, line 56: I would strongly suggest that you use the [bulk helper](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/client-helpers.html) instead of the raw API method - there are things it does under the hood around error checking and proper streaming that will be a pain to do ourselves.

api/test/fixtures/elastic.ts
```diff
@@ -0,0 +1,81 @@
+import { Client, errors } from "@elastic/elasticsearch";
```
jamieparkinson, line 1: Is this file being used at all? Since we're now getting stuff from Elasticsearch, I think we can use it to test our API endpoints, as we do in the concepts API.

jamieparkinson (commented):
This looks like a good start to me - I think what I'd suggest before merging is spending some time disentangling the Elasticsearch indexing stuff from the querying stuff and other API-related stuff, so it's easier to lift-and-shift the logic to the other app.

Mostly that means that the stuff in `elastic-search.ts` (nit: it's always elasticsearch, never elasticSearch or elastic-search or elastic_search etc!) is currently expecting to be used as part of an Express endpoint, which is not its final destination - perhaps you could split it up into functions for (eg) 1. getting stuff from Prismic,  2. transforming it, and 3. indexing it.

Once that's done, I think it'll be easy to tighten up the scope of this PR and split it: maybe 1 PR for adding functions for indexing things in to ES, and 1 for reading documents from ES.

api/src/helpers/index.ts
```diff
@@ -5,10 +5,6 @@ export function isNotUndefined<T>(val: T | undefined): val is T {
   return typeof val !== "undefined";
 }
 
-export function isUndefined<T>(val: T | undefined): val is T {
```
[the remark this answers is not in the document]
jamieparkinson: ha, totally wrong!

api/test/fixtures/elastic.ts
rcantin-w: No you're right we can delete it

PR #9, opened 2023-03-21
Add pipeline project skeleton

jamieparkinson:
The way the packaging works is copied from https://github.com/wellcomecollection/elasticsearch-log-forwarder, which seems to work quite well.


pipeline/package.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -eo pipefail
```
alexwlchan, line 2: nit: I prefer the longest versions of these, as easier to google and more readable to people who aren't super-familiar with bash

```suggestion
#!/usr/bin/env bash
set -o errexit
set -o nounset
set -o pipefail
```

pipeline/test/pipeline.test.ts
```diff
@@ -0,0 +1,5 @@
+describe("pipeline", () => {
+  it("exists", () => {
+    expect(2 + 2).toBe(4);
```
alexwlchan, line 3:
```
npm install peano-axioms
```

alexwlchan (approved):

pipeline/test/pipeline.test.ts
jamieparkinson: previously I did not manage to refrain from putting this terrible joke into code https://github.com/wellcomecollection/concepts-pipeline/blob/0ea38f4060058678b0d1c52ef0473d9839e97415/ingestor/src/test/scala/weco/concepts/ingestor/Main.scala#L6-L10

PR #15, opened 2023-03-23
WIP: Restructure pipeline

rcantin-w:


api/src/app.ts
```diff
@@ -1,21 +1,21 @@
...
+import { Client as ElasticClient } from "@elastic/elasticsearch";
 import {
   errorHandler,
   articlesController,
   articleController,
 } from "./controllers";
 import { Config } from "../config";
-import { Clients } from "./types";
 
-const createApp = (clients: Clients, config: Config) => {
```
jamieparkinson, line 12: Not especially important, but I would personally not have changed this - if we needed to access another external service at some point, we'd have to reintroduce the generic `Clients` as this has coupled us to an `app` using one single `ElasticClient`.

pipeline/config.ts
```diff
@@ -0,0 +1,19 @@
...
+  PUBLIC_ROOT_URL: z
+    .string()
+    .url()
+    .default("https://api.wellcomecollection.org/content/v0"),
+});
+const environment = environmentSchema.parse(process.env);
+
+const config = {
+  contentsIndex: "testing-index",
+  publicRootUrl: new URL(environment.PUBLIC_ROOT_URL),
```
jamieparkinson, line 14: Not needed in the pipeline (above as well)

pipeline/package.json
```diff
@@ -9,11 +9,18 @@
     "clean": "rimraf ./dist package.zip",
     "test": "jest"
   },
+  "dependencies": {
+    "@prismicio/client": "^6.7.3",
+    "@prismicio/helpers": "^2.3.9",
+    "@prismicio/types": "^0.2.7",
```
jamieparkinson, line 15: Should this be a dev dependency?

api/src/types/index.ts
```diff
@@ -1,46 +1,47 @@
-import * as prismic from "@prismicio/client";
-import { Client as ElasticClient } from "@elastic/elasticsearch";
-
-// Generic types
-export type ImageDimensions = {
-  width?: number;
-  height?: number;
+// TODO keep?
+import { ArticleFormatId, PrismicImage } from "../../../pipeline/src/types";
```
jamieparkinson, line 2: I think we probably want to remove this dependency between the 2 apps, and actually remove all of the types here - on reflection, it probably wasn't the right call in the Concepts API to have the return types represented in TypeScript, as the `display` document is intended to be completely opaque JSON (ie the API doesn't need to know its type). In other words, we don't need the API to know what an `Article` is, but we do need the pipeline to know.

pipeline/src/helpers/elasticsearch.ts
```diff
@@ -0,0 +1,57 @@
...
+type QueryParams = {
+  query?: string;
+} & PaginationQueryParameters;
+
+export type Handler = RequestHandler<
+  never,
+  { count: CountResponse },
+  never,
+  QueryParams
+>;
```
jamieparkinson, line 18: Can be removed, these are API things

pipeline/src/helpers/elasticsearch.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+type TransformedDocument = {
+  id: string;
+};
+
+export const bulkIndexDocuments = async (
+  elasticClient: Client,
+  docs: TransformedDocument[]
+) => {
```
jamieparkinson, line 39: fyi I would recommend doing this like so, to make the types as informative as possible
```typescript
type HasIdentifier = {
  id: string;
}

export const bulkIndexDocuments = async <T extends HasIdentifier>(elasticClient: Client, docs: T[]) => { ... }
```

jamieparkinson (commented):
Realise I might've made some comments on stuff that isn't new but has just been moved from the last PR  - if you'd prefer, feel free to address them in a separate PR

pipeline/src/handler.ts
```diff
@@ -1,4 +1,40 @@
...
+          .slice(0, 40)
+          .map((result) => transformArticle(result));
+
+        // Bulk send them to Elasticsearch
+        await bulkIndexDocuments(clients.elastic, transformedResponse);
+      } else {
+        throw new HttpError({
+          status: 404,
+          label: "No results found",
+        });
```
jamieparkinson, line 32: Can just throw a plain old fashioned error here, no need for the express stuff

pipeline/src/helpers/error.ts
```diff
@@ -0,0 +1,60 @@
+import { ErrorRequestHandler } from "express";
```
jamieparkinson, line 1: Can remove this whole file (and the express dependency) as per https://github.com/wellcomecollection/content-api/pull/15/files#r1147485796

pipeline/src/helpers/prismic.ts
```diff
@@ -36,10 +36,10 @@ export const getPrismicDocuments = async <T>({
   prismicClient: prismic.Client;
   contentTypes: ContentType[];
 }): Promise<T> => {
-  const getDocuments = await prismicClient.getAllByType(contentTypes[0], {
+  const getDocuments = await prismicClient.getByType(contentTypes[0], {
```
rcantin-w, line 39: Wanted to simplify, don't need to get all 900 articles for this

jamieparkinson (approved):

PR #16, opened 2023-03-24
Elasticsearch cluster in Terraform

jamieparkinson:
Primarily cribbed from the concepts API and pipeline services.

- Creates a new cluster, `content-2023-03-24`
- Creates users and roles for the api and pipeline services
- Switches the API to use those new roles


api/server.ts
```diff
@@ -8,10 +8,12 @@ import { getElasticClient } from "./src/services/elasticsearch";
...
-  app.listen(port, () => {
-    log.info(`Content API listening on port ${port}`);
-  });
-});
+getElasticClient({ pipelineDate: config.pipelineDate }).then(
+  async (elasticClient) => {
+    const app = createApp({ elastic: elasticClient }, config);
+    const port = process.env.PORT ?? 3000;
+    app.listen(port, () => {
+      log.info(`Content API listening on port ${port}`);
```
alexwlchan, line 16: minor: I find it useful to make this a link that goes directly to a working page, e.g.

```suggestion
      log.info(`Content API listening on https://localhost:${port}/articles`);
```

alexwlchan (approved):

PR #17, opened 2023-03-29
Move stuff into a common project

jamieparkinson:


pipeline/package.json
```diff
@@ -26,6 +27,6 @@
     "rimraf": "^4.4.0",
     "ts-jest": "^29.0.5",
     "ts-node": "^10.9.1",
-    "typescript": "^5.0.2"
+    "typescript": "^4.9.5"
```
rcantin-w, line 30: just curious why?

rcantin-w (approved):

pipeline/package.json
jamieparkinson: ts-jest was complaining about compatibility with 5, looks like they've now released support but will deal with that separately!

PR #27, opened 2023-04-03
Add sorting by date

rcantin-w:
`?sort=publication.date&sortOrder=asc`, for example.
Defaults to `_score`. If no `sortOrder` is specified, defaults to `desc`. 

As specified in the RFC, only sorting by date or relevance.

### DIFF
--- a/api/src/controllers/articles.ts
+++ b/api/src/controllers/articles.ts
@@ -17,6 +17,8 @@ type QueryParams = {
 
 type ArticlesHandler = RequestHandler<never, ResultList, never, QueryParams>;
 
+const articlesSortValues = ["publication.date"];
+
 const articlesController = (
   clients: Clients,
   config: Config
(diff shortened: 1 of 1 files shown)


api/src/controllers/articles.ts
```diff
@@ -53,6 +55,16 @@ const articlesController = (
               }
             : undefined,
         },
+        sort:
+          sortBy === "publication.date"
```
jamieparkinson, line 59: Should we return a 400 if it's not a value we expect? cf. https://api.wellcomecollection.org/catalogue/v2/works?sort=BLAH

rcantin-w: Good idea. [Done here](https://github.com/wellcomecollection/content-api/pull/27/commits/7bb1b362c979109e284399410a23d46fee26e521). Might be slightly overkill but wanted it to be readable and flexible.

rcantin-w: should I also force the `sortOrder` if `sort` is specified or is that alright?

jamieparkinson (approved):

PR #30, opened 2023-04-11
Set a default format of Article

rcantin-w:
Some articles don't have a defined format, in which case we always want it to be Article.

Twinned with [wc.org changes](https://github.com/wellcomecollection/wellcomecollection.org/pull/9577)

Relates to #9575

### DIFF
--- a/pipeline/src/transformers/article.ts
+++ b/pipeline/src/transformers/article.ts
@@ -16,6 +16,7 @@ import {
   EmptyImageFieldImage,
   PrismicDocument,
 } from "@prismicio/types";
+import { defaultArticleFormat } from "@weco/content-common/data/formats";
 
 const getContributors = (
   document: PrismicDocument<WithContributors>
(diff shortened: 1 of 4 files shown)


pipeline/src/transformers/article.ts
```diff
@@ -101,7 +101,11 @@ export const transformArticle = (
 
   const format = isFilledLinkToDocumentWithData(data.format)
     ? transformLabelType(data.format)
-    : undefined;
+    : ({
+        type: "ArticleFormat",
+        id: "W7TfJRAAAJ1D0eLK",
+        label: "Article",
+      } as ArticleFormat);
```
alexwlchan, line 108: Minor: when we're hard-coding formats like this, maybe put them in a top-level directory similar to `data` in the front-end repo? So we can see all the hard-coded data in one place.

alexwlchan (approved):

PR #33, opened 2023-04-11
index the complete article body

harrisonpim:
At the moment we're not indexing body text beyond the first paragraph of an article's body, so a lot of documents with relevant keywords in their body are being missed! For example, [_Are doctors medical detectives?_](https://wellcomecollection.org/articles/XKIXNRAAAJsnbmFE) is missing from a search for **parkinsons** because the word only appears after the second paragraph.

This PR indexes _all_ of the body's `text` fields, and adds in `quoteV2`s and `standfirst`s for good measure.

I've also shuffled a couple of the filters in the mapping - these should only have a minor positive effect on scoring.

### DIFF
--- a/pipeline/src/indices/articles.ts
+++ b/pipeline/src/indices/articles.ts
@@ -115,13 +115,9 @@ export const settings = {
         type: "stemmer",
         language: "possessive_english",
       },
-      asciifolding_token_filter: {
-        type: "asciifolding",
-        preserve_original: true,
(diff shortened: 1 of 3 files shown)


rcantin-w (approved):
nice one! that's been reindexed right? (looks good in console)

harrisonpim:
Yep, successfully reindexed last night. Thanks for the instructions!

PR #35, opened 2023-04-13
Use PrivateLink again

jamieparkinson:
I was wrong about this being a bug in the ES client - the issue was that (a) I'd forgotten to configure the PrivateLink traffic filters for the cluster, and (b) the way we chose which endpoint to use was indirect/unclear and think had changed inadvertently, obscuring the actual issue and leading me on a wild goose chase.

### DIFF
--- a/common/services/elasticsearch.ts
+++ b/common/services/elasticsearch.ts
@@ -14,9 +14,7 @@ const getElasticClientConfig = async ({
   hostEndpointAccess = "public",
 }: ClientParameters): Promise<ClientOptions> => {
   const secretPrefix = `elasticsearch/content-${pipelineDate}`;
-  if (hostEndpointAccess === "private") {
-    log.info("Creating ES client for PrivateLink endpoint");
-  }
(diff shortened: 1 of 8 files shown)


common/services/elasticsearch.ts
```diff
@@ -1,21 +1,24 @@
...
 
 const getElasticClientConfig = async ({
   pipelineDate,
   serviceName,
+  hostEndpointAccess = "public",
 }: ClientParameters): Promise<ClientOptions> => {
   const secretPrefix = `elasticsearch/content-${pipelineDate}`;
+  if (hostEndpointAccess === "private") {
+    log.info("Creating ES client for PrivateLink endpoint");
+  }
```
alexwlchan, line 19: Is it worth adding a log line for the public endpoint, so we'll know if it's doing the wrong thing from a log that's there, not a log that isn't?

alexwlchan (approved):

common/services/elasticsearch.ts
jamieparkinson: Yes

PR #40, opened 2023-04-21
Querying series title and contributors for articles/webcomics

jamieparkinson:
Gets series titles and contributors from Prismic, indexes them alongside articles/webcomics, updates the API to query them. Diff is a bit bigger than one might hope as I've moved some stuff around.


harrisonpim:
Can we add the article IDs to the queryable fields, alongside the new series IDs? I overlooked article IDs in the original mapping but they're now here as [keywords in the prototype mapping](https://github.com/wellcomecollection/data-science/blob/3a244941847f6e63f46c17b0a47470a10cca2374/prismic-pipeline-prototype/data/index_config/articles.json#L4-L6) and in [the query](https://github.com/wellcomecollection/data-science/blob/3a244941847f6e63f46c17b0a47470a10cca2374/prismic-pipeline-prototype/data/queries/articles.json#L6)

jamieparkinson:
Yes, will do in a separate PR so as not to pollute the intent of this one

rcantin-w (approved):
From what I can see it all makes a lot of sense! Excited to see that 👀

PR #47, opened 2023-05-10
Reflect updates to linked documents

jamieparkinson:
What this PR contains:
- A function for extracting all linked document IDs from arbitrary objects
- An update to the transformer/mapping which finds linked document IDs inside articles and indexes them
- An implementation of the logic proposed in [RFC 056](https://github.com/wellcomecollection/docs/tree/main/rfcs/056-prismic-etl-pipeline#proposal) such that if a child document is updated, the pipeline Lambda finds any parent articles in which it is used and updates them too.
- Tests for the above!


pipeline/test/elasticsearch-helpers.test.ts
```diff
@@ -52,3 +58,32 @@ describe("ensureIndexExists", () => {
...
+        getParentDocumentIDs(testClient, {
+          index: "test",
+          identifiersField: "childIds",
+          batchSize: batchSize,
+        })
+      )
+    );
+
+    expect(elasticScrollDocuments).toHaveBeenCalledTimes(totalDocs / batchSize);
+    expect(finalDoc).toBe(totalDocs.toString());
```
paul-butcher, line 87: At first glance, I found this a little confusing (though a closer look clarified it)  - why is finalDoc the same value as totalDocs?  Is it picking out a count or an id?

I wonder if it might be clearer if identifiedDocuments were to create an id based on the number, rather than just stringifying it (e.g. "id_100")

jamieparkinson: ah yes, I was being lazy here and I think had lied to myself that I'd come back to it - will sort it

pipeline/src/handler.ts
```diff
@@ -20,34 +30,58 @@ export const createHandler =
...
+      // The field name is mapped in `indices/articles.ts` and populated by the transformer
+      getParentDocumentIDs(clients.elastic, {
+        index: articles.index,
+        identifiersField: "query.linkedIdentifiers",
+      }),
+      // We don't need to update parent documents that we already got in this window
+      // as they were updated above
+      filter((parentId) => !initialIndex.successfulIds.has(parentId)),
+      // Fetch the latest version of all the parent documents
+      getDocumentsByID<ArticlePrismicDocument>(clients.prismic, { graphQuery })
```
paul-butcher, line 76: IIUIC, this fetches the parent document with its child content denormalised into it, meaning that we don't have to do any fiddling after this to insert the child content.

Could something along those lines be added to the comment?  It would help with understanding both the why and the where. (Why don't we just get it from ES, since we're doing the IDs query, above?  Where do we populate the parent with the new data?)

pipeline/src/handler.ts
```diff
@@ -20,34 +30,58 @@ export const createHandler =
...
+          graphQuery,
+          after,
+        })
+      ),
+      isArticle
+    );
+
+    // 3. Index all the parent documents into elasticsearch after transforming them
+    // into our common schema
+    const initialIndex = await bulkIndexDocuments(
```
paul-butcher, line 53: Does the two-stage indexing thing do something important?

IIUIC, bulkIndexDocuments now throws when a single document fails, so if any of these top-level documents fails to index (possibly even for transient reasons), then none of the nextParentDocuments will be updated.

Could the indexing of all of them be done in one call?

paul-butcher (approved):

pipeline/src/handler.ts
jamieparkinson: It did but I have now removed that - the last 3 commits should explain the situation. Re the throwing-on-failure, I decided that doing that (so we get a Slack alert) was the easiest approach for now!

Issue #48, opened 2023-05-16
Content API faceting


jamieparkinson:
https://wellcome.slack.com/archives/C02ANCYL90E/p1684148576908159

PR #51, opened 2023-05-23
Add aggregations with correct faceting behaviour

jamieparkinson:
Big diff *but* it's mostly tests and comments! 😇 

This PR adds aggregations to the content API. The complexity of the change is due to the need to return results that can be used to build a [faceted](https://en.wikipedia.org/wiki/Faceted_search) search interface. The details of those requirements are explained in [RFC 037](https://github.com/wellcomecollection/docs/tree/main/rfcs/037-api-faceting-principles).

The Catalogue API achieves this behaviour through many classes spread across the application, and uses the https://github.com/sksamuel/elastic4s DSL to construct the queries. This code is extremely complex, confusing, and poorly understood (I can say that because I wrote it!)

Most of that code was written during the process of understanding the desired behaviour, rather than after we had a good understanding of requirements: this PR isn't a port of that code to TypeScript but rather it's a rewrite given the understanding we have now. 

A major difference is that I have [reverted](https://github.com/wellcomecollection/catalogue-pipeline/pull/1520) the change that was made in the Catalogue API where we switched from `post_filter` and `min_doc_count: 0` to using the `global` aggregation. It's questionable whether the latter was ever more performant and I think the `post_filter` approach is a clearer statement of intent, and makes for simpler code all round. I also discovered while reading the docs that we can get ES to sort by sug-aggregation for us, removing the need for [reordering](https://github.com/wellcomecollection/catalogue-api/blob/main/search/src/main/scala/weco/api/search/models/Aggregation.scala#L102-L104) responses in the application.

**Interesting bits**
- [Transforming aggregations from ES into our response format](https://github.com/wellcomecollection/content-api/pull/51/files#diff-6bf5f20eed98b41fca5854dac3a9d980e781f811ad40dc011a2002d0faacb15bR12-R42)
- [New fields in the index to aggregate upon](https://github.com/wellcomecollection/content-api/pull/51/files#diff-11bb4100206da7bc8f2c551c3831b624c118128a95951e5427fc79710d76fd27R184-R187)
- [Logic for rewriting standalone aggregations and filters to fulfil facet requirements](https://github.com/wellcomecollection/content-api/pull/51/files#diff-5fd53a7634e25318abd217a9a2171f5abafd4cae08fd5f704757e3b3ba215f71)
- [The actual aggregations we're using](https://github.com/wellcomecollection/content-api/pull/51/files#diff-415d82a7d58b2c8b279d6a6f5e53fee1c4e98e8792c5a072cc264ea423af5f46R48-R66)

### DIFF
--- a/api/src/controllers/articles.ts
+++ b/api/src/controllers/articles.ts
@@ -78,6 +78,7 @@ const articlesController = (
       articlesFilter
     );
 
+    // See comments in `queries/faceting.ts` for some explanation of what's going on here
     const facetedAggregations = ifDefined(initialAggregations, (aggs) =>
       rewriteAggregationsForFacets(aggs, initialFilters)
     );
--- a/api/src/controllers/articles.ts
+++ b/api/src/controllers/articles.ts
@@ -15,6 +15,11 @@ import { HttpError } from "./error";
 import { ResultList } from "../types/responses";
 import { resultListResponse } from "../helpers/responses";
 import { AggregationsAggregate } from "@elastic/elasticsearch/lib/api/types";
+import {
+  partitionFiltersForFacets,
+  rewriteAggregationsForFacets,
+} from "../queries/faceting";
+import { pickFiltersFromQuery } from "../helpers/requests";
 
 type QueryParams = {
   query?: string;
@@ -64,6 +69,26 @@ const articlesController = (
     const sortKey =
       sort === "publicationDate" ? "query.publicationDate" : "_score";
 
+    const initialAggregations = ifDefined(aggregations, (requestedAggs) =>
+      pick(articlesAggregations, requestedAggs)
+    );
+    const initialFilters = pickFiltersFromQuery(
+      ["contributors.contributor", "format"],
+      params,
+      articlesFilter
+    );
+    const { postFilters, queryFilters } = partitionFiltersForFacets(
+      initialAggregations ?? {},
+      initialFilters
+    );
+    const dateFilter =
+      params["publicationDate.from"] || params["publicationDate.to"]
+        ? articlesFilter.publicationDate(
+            ifDefined(params["publicationDate.from"], validateDate),
+            ifDefined(params["publicationDate.to"], validateDate)
+          )
+        : undefined;
+
     try {
       const searchResponse = await clients.elastic.search<
         Displayable,
@@ -73,25 +98,18 @@ const articlesController = (
       >({
         index,
         _source: ["display"],
-        aggregations: ifDefined(aggregations, (requestedAggs) =>
-          pick(articlesAggregations, requestedAggs)
+        aggregations: ifDefined(initialAggregations, (aggs) =>
+          rewriteAggregationsForFacets(aggs, initialFilters)
         ),
         query: {
           bool: {
             must: ifDefined(queryString, articlesQuery),
-            filter: [
-              ifDefined(
-                params["contributors.contributor"]?.split(","),
-                articlesFilter["contributors.contributor"]
-              ),
-              ifDefined(params.format?.split(","), articlesFilter.format),
-              params["publicationDate.from"] || params["publicationDate.to"]
-                ? articlesFilter.publicationDate(
-                    ifDefined(params["publicationDate.from"], validateDate),
-                    ifDefined(params["publicationDate.to"], validateDate)
-                  )
-                : undefined,
-            ].filter(isNotUndefined),
+            filter: [...queryFilters, dateFilter].filter(isNotUndefined),
+          },
+        },
+        post_filter: {
+          bool: {
+            filter: postFilters,
           },
         },
         sort: [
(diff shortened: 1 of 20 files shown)


alexwlchan:
Jamie and I just had a call to go through this together; here's a list of actions:

* [ ] JP: Review all the code around empty buckets, which includes:

    * Why are we filtering out empty buckets in the front end?
    * Removing buckets which (1) have zero results and (2) don't have a currently applied filter
    
    As part of this, I recalled writing some code to deal with filters/aggregations. I think that was https://github.com/wellcomecollection/wellcomecollection.org/pull/8513, which includes a comment I now believe to be wrong _"We build the list of available filters from […] the list of values the user has already selected, which may not appear in the aggregations"_

* [ ] AC: Document the reason we aren't doing aggregations after the initial query (because we need to accommodate no-JS users)
* [ ] AC: Add some concrete examples to the RFC about API faceting, including a link to the Elasticsearch docs that Jamie has found
* [ ] JP: Make `dateFilters` a list and flatten it
* [ ] AC: Read the code on their own

alexwlchan:
part of https://github.com/wellcomecollection/content-api/issues/48; reviewing now

api/src/controllers/validation.ts
```diff
@@ -1,51 +1,92 @@
...
-import { isInArray } from "../helpers";
+import { isInSet, not } from "../helpers";
+import { StringLiteral } from "../types";
 
-type StringLiteral<T> = T extends string
-  ? string extends T
-    ? never
-    : T
-  : never;
+type NonEmptyArray<T> = [T, ...T[]];
```
alexwlchan, line 5: TIL!

(I've wondered for a while why Prismic's `RichTextField` type is defined as `[] | [RTNode, ...RTNode[]]`, and now it makes sense.)

api/src/controllers/validation.ts
```diff
@@ -1,51 +1,92 @@
...
+    const validValues = providedValues.filter(isInSet(allowedSet));
+    if (
+      validValues.length !== providedValues.length ||
+      validValues.length === 0
+    ) {
+      const invalidValues = providedValues.filter(not(isInSet(allowedSet)));
+      const invalidMessage =
+        invalidValues.length === 1
+          ? "is not a valid value"
+          : "are not valid values";
```
alexwlchan, line 37: nice

api/src/controllers/validation.ts
```diff
@@ -1,51 +1,92 @@
...
+    const providedValues = query[name]?.split(",");
+    if (providedValues === undefined) {
+      return defaultValue === undefined ? undefined : [defaultValue];
+    }
+
+    const validValues = providedValues.filter(isInSet(allowedSet));
+    if (
+      validValues.length !== providedValues.length ||
+      validValues.length === 0
+    ) {
```
alexwlchan, line 32: I see what you mean about missing Scala; this is the sort of thing where `.partition()` would come in really handy.

api/src/controllers/validation.ts
```diff
@@ -1,51 +1,92 @@
...
+      )} is not a valid date. Please specify a date or datetime in ISO 8601 format.`,
     });
   }
   return date;
 };
+
+const quoted = (str: string) => `'${str}'`;
+
+// For listing items (no Oxford comma)
+const readableList = (
```
alexwlchan, line 79: nit: not sure if this belongs in a controllers file or would be better pushed into some sort of grammar utils file (cf the front-end apps)

api/src/helpers/responses.ts
```diff
@@ -0,0 +1,81 @@
...
+              {
+                buckets: aggregation.buckets.map((bucket) => ({
+                  // This should always be a JSONified string, so if we can't
+                  // parse it something has gone wrong in the pipeline and we
+                  // should know about it
+                  data: JSON.parse(bucket.key),
+                  // If there is a filter subaggregation (named `filtered`), we should
+                  // use that: it will exist if other filters and aggregations are
+                  // applied in addition to the aggregation/filter corresponding to this bucket.
+                  count: bucket.filtered?.doc_count ?? bucket.doc_count,
```
alexwlchan, line 33: helpful comments 👍

api/src/controllers/articles.ts
```diff
@@ -29,49 +38,86 @@ const sortValidator = queryValidator({
...
+    );
+
+    // See comments in `queries/faceting.ts` for some explanation of what's going on here
+    const facetedAggregations = ifDefined(initialAggregations, (aggs) =>
+      rewriteAggregationsForFacets(aggs, initialFilters)
+    );
+    const { postFilters, queryFilters } = partitionFiltersForFacets(
+      initialAggregations ?? {},
+      initialFilters
+    );
```
alexwlchan, line 88: as discussed yesterday, I 💚 the choice of variable names here

alexwlchan (approved):

alexwlchan (approved):

api/src/controllers/validation.ts
jamieparkinson: Yeah once you see it it's quite an elegant way of expressing this imo!

api/src/controllers/validation.ts
jamieparkinson: At some point I predict that these utils will be commonised (ie `typescript-libs`) so will deal with that then

PR #55, opened 2023-06-07
Correct understanding of faceting

jamieparkinson:
I was _rather pleased with myself_ after https://github.com/wellcomecollection/content-api/pull/51 - but I shouldn't have been, I got it wrong. The test cases should adequately illustrate what the misunderstanding was...


jamieparkinson:
Github has done something weird with this PR

PR #56, opened 2023-06-07
Correct understanding of faceting

jamieparkinson:
I was rather pleased with myself after https://github.com/wellcomecollection/content-api/pull/51 - but I shouldn't have been, I got it wrong. The test cases should adequately illustrate what the misunderstanding was...


georgiaewhitney:
face-inating

georgiaewhitney (approved):
my rough understanding is that filtering should be more accurate/optimised as now modified to include post-filtering to each aggregation, excluding its initial filter and providing more precise results

PR #73, opened 2023-11-03
Events graph query and snapshots

agnesgaroux:
[Issue](https://github.com/wellcomecollection/content-api/issues/63)


rcantin-w:
`availableOnline` would probably used for filtering as well as possible labelling (We currently use it to add "Available to watch")

Screenshot 2023-11-03 at 13 42 16

`promo` is where we get our image from, it's where they put the Editorial image and its alt text.

You can look at transformers for more detail, I might be missing other uses for the above. [See here for the generic fields](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/prismic/transformers/index.ts#L120) transformer

jamieparkinson (approved):
v nice

Issue #64, opened 2023-10-27
Content API/Events: Build transformer function


rcantin-w:
- The [article transformer](https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/transformers/article.ts#L0-L1) has a few helpful functions (e.g. getContributors) that could be moved to a shared file.
- At this point, only focus on having an `id` and a `display` object in the returned object, `query` etc. can be done later
- Ensure everything is typed; new types will need to be defined but a lot of the Prismic ones should be available already.

PR #75, opened 2023-11-21
Event documents transformer

agnesgaroux:
[Issue](https://github.com/wellcomecollection/content-api/issues/64) 

- added contributors to the eventDoc query: that makes its type the same as series on articles
- ~the above could actually be removed, since~ I ended up not including `contributors` in `display` or `query`. Reason is, it doesn't seem important enough to display on the result list, and I don't see people querying for specific hosts or facilitator. Happy to discuss 💬
- Any idea what we could use as "default event format"? it's sometimes undefined 
- not sure about the `times` in the `display`


rcantin-w:
> the above could actually be removed, since I ended up not including contributors in display or query. Reason is, it doesn't seem important enough to display on the result list, and I don't see people querying for specific hosts or facilitator. Happy to discuss 💬

I thiiiink this was because of curators or specific artists, but tbh I'm not sure! Happy to leave out for now and add later if we want it back?

pipeline/src/transformers/eventDocument.ts
```diff
@@ -0,0 +1,131 @@
...
+  isImageLink,
+  asText,
+  asTitle,
+  isNotUndefined,
+} from "../helpers/type-guards";
+import { linkedDocumentIdentifiers, formatSeriesForQuery } from "./utils";
+
+const onlineLocation = {
+  type: "EventLocation",
+  id: "ef04c8e3-26be-4fbc-9bef-f52589ebc56c",
```
rcantin-w, line 25: Where is this id from? (not Prismic?)

pipeline/src/types/transformed/eventDocument.ts
```diff
@@ -0,0 +1,30 @@
+import { Image } from ".";
+
+export type EventDocument = {
+  type: "Event";
+  id: string;
+  title: string;
+  image?: Image;
+  times: { startDateTime: Date | null; endDateTime: Date | null }[];
+  format: EventDocumentFormat | undefined;
```
rcantin-w, line 9: Could we do `format?: EventDocumentFormat;` to match the others?

pipeline/src/types/transformed/eventDocument.ts
```diff
@@ -0,0 +1,30 @@
+import { Image } from ".";
+
+export type EventDocument = {
+  type: "Event";
+  id: string;
+  title: string;
+  image?: Image;
+  times: { startDateTime: Date | null; endDateTime: Date | null }[];
```
rcantin-w, line 8: Can we do the same here or does it have to be `null`?

pipeline/src/transformers/eventDocument.ts
agnesgaroux: As discussed, isOnline is a boolean in Prismic. We want to treat it as a location so it needs to be in the same format as physical location

pipeline/src/types/transformed/eventDocument.ts
agnesgaroux: Aligning on how the FE handles this, it is now optional to account for the possibility of it being undefined (instead of null)

common/data/defaultValues.ts
```diff
@@ -0,0 +1,18 @@
+// used for display purpose only, when the prismic data is undefined
+export const defaultArticleFormat = {
+  type: "ArticleFormat",
+  id: "W7TfJRAAAJ1D0eLK",
+  label: "Article",
+};
+
+export const defaultEventFormat = {
+  type: "EventFormat",
+  id: "dfc2b7f9-c362-47da-9644-b0f98212ccaa",
```
rcantin-w, line 10: If we're keeping this I'd be tempted to add a comment to explain the difference in IDs

pipeline/src/graph-queries/eventDocuments.ts
```diff
@@ -61,6 +61,16 @@ const query = `{
     series {
       series {
         title
+        contributors {
```
rcantin-w, line 64: Is this what it related to? 
https://github.com/wellcomecollection/content-api/pull/75#issuecomment-1824118943

rcantin-w (approved):

pipeline/src/graph-queries/eventDocuments.ts
agnesgaroux: Ha! no I got the original commit comment wrong (fixed it now) because I confused contributors and series contributors
The EventDocument query gets:
- Contributors, which I left out of the transformed document because it doesn't seem very useful for either display or query
- Series, which now includes Series contributors, so it's aligned with Articles

rcantin-w: So we're fetching Contributors but not using them afterwards at the moment?

common/data/defaultValues.ts
agnesgaroux: ✅

pipeline/src/graph-queries/eventDocuments.ts
agnesgaroux: Indeed. The idea is the query has pretty much everything we _might_ want to use, we can trim it down once the transformed type is a bit more stable/developed to remove anything we don't actually want in the index

rcantin-w: Works for me :)

Issue #68, opened 2023-10-27
Single event endpoint


jamieparkinson:
**Phase 1**

We want to be able to `GET /events/:id` for a single event. This will follow the prior art of the `controller/article` etc in `/api/src`.

A suggested work order would be:

- Minor changes so the API expects to be dealing with multiple indices: `config.ts` to be changed from having a single `contentsIndex` to having an `articlesIndex` and an `eventsIndex` or something like that.
- An `event.ts` controller similar to `article.ts`
- An `/event/:id` route in `app.ts` - would expect this to 404 when queried since there's no index.
- Put some test data in the index (see below)
- An `event.test.ts` equivalent to `article.test.ts`

**Running the API locally**
You can run the API locally with auto-reload on code changes with `yarn dev` in `/api`.

If you just do this it will complain about not being able to fetch the secrets to talk to Elasticsearch, so tell it what profile to use: `AWS_PROFILE=catalogue-developer yarn dev` (and make sure you have a recent token ie `aws-azure-login`).

**Test data**
If you open the content cluster in Elastic Cloud and go to `Dev Tools` (bottom of the burger menu) you can...

- Create an index with `PUT /events-test` (`-test` suffix so as not to collide with the pipeline lambda)
- Create a test document with a display model that looks vaguely like [this](https://github.com/wellcomecollection/docs/blob/main/rfcs/055-prismic-api/events.md#example-response):
```
POST /events-test/_doc/<PUT AN ID HERE>
{
  "display": { copy something out of the RFC here }
}
```

PR #76, opened 2023-12-01
Single event endpoint

georgiaewhitney:
### As detailed in #68

- The content-api now expects multiple indices:
  - `contentsIndex` is now split into `articlesIndex` and `eventsIndex`
---

- There is a controller for events, following the same article logic:
  - `controllers/event.ts`

---

- A route for `/event:id` is in the `app.ts`:
  - a call to `/events/:id` correctly 404s with no index yet in place

```json
{
  "type": "Error",
  "httpStatus": 404,
  "label": "Not Found",
  "description": "Event not found for identifier xx",
  "errorType": "http"
}
```

---

- A test data index is in place (requires config to be adjusted to call `"events-test"`) which returns a test event under id: `abc123`:
```json
{
  "type": "Event",
  "id": "abc123",
  "title": "Test Event",
  "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat."
}
```
---

- A test similar to `article.test.ts` is in place and passes for events:
  - `event.test.ts` returns a document for the given id and returns a 404 if no document for the id exists

### DIFF
--- a/api/config.ts
+++ b/api/config.ts
@@ -11,7 +11,8 @@ const environment = environmentSchema.parse(process.env);
 
 const config = {
   pipelineDate: "2023-03-24",
-  contentsIndex: "articles",
+  articlesIndex: "articles",
+  eventsIndex: "events",
   publicRootUrl: new URL(environment.PUBLIC_ROOT_URL),
 };
 
--- a/api/config.ts
+++ b/api/config.ts
@@ -11,7 +11,9 @@ const environment = environmentSchema.parse(process.env);
 
 const config = {
   pipelineDate: "2023-03-24",
-  contentsIndex: "articles",
+  articlesIndex: "articles",
+  eventsIndex: "events",
+  // contentsIndex: "articles",
   publicRootUrl: new URL(environment.PUBLIC_ROOT_URL),
 };
 
--- a/api/config.ts
+++ b/api/config.ts
@@ -12,8 +12,10 @@ const environment = environmentSchema.parse(process.env);
 const config = {
   pipelineDate: "2023-03-24",
   articlesIndex: "articles",
-  eventsIndex: "events",
-  // contentsIndex: "articles",
+  eventsIndex: "events-test",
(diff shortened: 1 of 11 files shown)


api/config.ts
```diff
@@ -11,7 +11,11 @@ const environment = environmentSchema.parse(process.env);
 
 const config = {
   pipelineDate: "2023-03-24",
-  contentsIndex: "articles",
+  articlesIndex: "articles",
+  eventsIndex: "events",
+  // eventsIndex: "events-test"
+  // test data with data under id abc123
+  // use GET /events/abc123
```
jamieparkinson, line 18: Fine to delete this comment now

api/src/controllers/error.ts
```diff
@@ -47,7 +47,7 @@ export const errorHandler: ErrorRequestHandler = (err, req, res, next) => {
     res.status(err.status).json(err.responseJson);
   } else {
     // Log this to prevent it getting swallowed
-    log.error(err);
+    console.error(err);
```
jamieparkinson, line 50: Can get rid of this now after syncing with main, as was fixed by https://github.com/wellcomecollection/content-api/pull/74/files

api/test/fixtures/api.ts
```diff
@@ -63,7 +63,8 @@ export const mockedApi = <T extends Displayable & Identified>(
     },
     {
       pipelineDate: "2222-22-22",
-      contentsIndex: testIndex,
+      articlesIndex: testIndex,
+      eventsIndex: testIndex,
```
jamieparkinson, line 67: To prevent a future headache might be worth splitting out a `testArticlesIndex` and a `testEventsIndex` here

jamieparkinson (approved):
🎉

Issue #65, opened 2023-10-27
Make the pipeline Lambda able to handle multiple document types


rcantin-w:
**Phase 1**

[This function](https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/handler.ts#L23) currently only considers articles. This should become more flexible and eventually allow for all kinds of content types (not just articles and events).

More functions across will probably require the change but this is a starting point.

PR #78, opened 2023-12-05
Handler fetch and index event documents

agnesgaroux:
[Ticket](https://github.com/wellcomecollection/content-api/issues/65)


jamieparkinson (commented):
There is a lot of duplication of logic in this - could we write one ETL function which is then parametrised for each document type?

PR #79, opened 2023-12-07
Upgrade prismic package to match wc.org version

rcantin-w:
With the recent issues in the wc.org repo with the Prismic Image type changing, we thought it best to ensure types and package versions matched. See https://github.com/wellcomecollection/wellcomecollection.org/pull/10463

Tests all run fine, I don't know if there'd be anything else to check?

Relates to #80


pipeline/test/prismic-snapshots/WqfhAyUAAM8VVngg.events.json
agnesgaroux, line 1: If you've removed `WqfhAyUAAM8VVngg` (the image-less prismic doc) from `update-prismic-snapshots.ts` can you delete the prismic json and re-reun `jest update snapshots?

agnesgaroux (approved):

pipeline/test/prismic-snapshots/WqfhAyUAAM8VVngg.events.json
rcantin-w: Oh oops, good catch! I had rerun `update-prismic-snapshots` thinking it would delete it as part of the script, but didn't realise the jest update was also needed.  Have done so now!

Issue #67, opened 2023-10-27
Events listing endpoint


jamieparkinson:
**Phase 1** 

Following on from https://github.com/wellcomecollection/content-api/issues/68 with an controller/test/etc `tests`, equivalent to `articles`. 

You should probably start off with just a paginated list, then we can work on adding a simple query.

pollecuttn:
@georgiaewhitney I don't have permission to close this - can you do so, please.

georgiaewhitney:
> @georgiaewhitney I don't have permission to close this - can you do so, please.

Ah apologies - only just spotted this!

PR #81, opened 2023-12-15
Endpoint for listing all /events 

georgiaewhitney:
Following on from #68, this covers #67 

- An `eventsController` in `controller/events.ts` has been created to request, fetch, and list all events. 

image

Using the test data from #68 - results page looks like: 

```JSON
{
  "type": "ResultList",
  "results": [
    {
      "type": "Event",
      "id": "abc123",
      "title": "Test Event",
      "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat."
    },
    {
      "type": "Event",
      "id": "def456",
      "title": "Test Event 2 - Multiple",
      "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat."
    }
  ],
  "pageSize": 10,
  "totalPages": 1,
  "totalResults": 2
}
```

---

- A test was created in `events.test.ts` to check GET functionality for `/events` to ensure that the endpoint correctly returns a list of event data.

---

- `app.ts` was accordingly updated to route to `/events`

### DIFF
--- a/api/src/controllers/events.ts
+++ b/api/src/controllers/events.ts
@@ -4,23 +4,9 @@ import asyncHandler from "express-async-handler";
 import { Clients, Displayable } from "../types";
 import { PaginationQueryParameters, paginationElasticBody } from "./pagination";
 import { Config } from "../../config";
-import {
-  articlesAggregations,
-  articlesFilter,
-  articlesQuery,
-} from "../queries/articles";
-import { queryValidator, validateDate } from "./validation";
-import { ifDefined, pick } from "../helpers";
 import { HttpError } from "./error";
 import { ResultList } from "../types/responses";
 import { resultListResponse } from "../helpers/responses";
-import { AggregationsAggregate } from "@elastic/elasticsearch/lib/api/types";
-import {
-  partitionFiltersForFacets,
-  rewriteAggregationsForFacets,
-} from "../queries/faceting";
-import { esQuery } from "../queries/common";
-import { pickFiltersFromQuery } from "../helpers/requests";
 
 type QueryParams = {
   query?: string;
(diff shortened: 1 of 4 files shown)


api/src/controllers/events.ts
```diff
@@ -1 +1,52 @@
...
+  return asyncHandler(async (req, res) => {
+    const { query: queryString, ...params } = req.query;
+    try {
+      const searchResponse = await clients.elastic.search<Displayable>({
+        index,
+        _source: ["display"],
+      });
+
+      res.status(200).json(resultList(req, searchResponse));
+    } catch (error) {
```
rcantin-w, line 37: This is all very far in my memory, but looking at the `articles` controller, there is [a little bit more done in the error handling](https://github.com/wellcomecollection/content-api/blob/main/api/src/controllers/articles.ts#L136) which might be useful to this endpoint as well. 

And remind me, I think it was said in a meeting, are we skipping pagination for now?

api/src/controllers/events.ts
```diff
@@ -1 +1,52 @@
...
+type QueryParams = {
+  query?: string;
+  sort?: string;
+  sortOrder?: string;
+  aggregations?: string;
+  "contributors.contributor"?: string;
+  "publicationDate.from"?: string;
+  "publicationDate.to"?: string;
+  format?: string;
+} & PaginationQueryParameters;
```
rcantin-w, line 20: Maybe more for @agnesgaroux to review the list, I think we were skipping contributors for now? I know we're doing a very generic version of it all for now, so might be a nitpick.

api/src/controllers/events.ts
georgiaewhitney: Yes! Will get this a bit more fleshed out now to match `articles` better. 

I thought we were doing pagination - I've just realised this isn't working.

api/src/controllers/events.ts
georgiaewhitney: Not a nitpick - is a valid point! Can remove them for now :)

rcantin-w: Just compare with what's currently there on the work Agnès did I reckon

agnesgaroux: I started thinking about filters when I was building the EventDocument transformer but realised it wasn't really a dev decision, so we don't have aggregations or filters in event documents yet
I'd just keep 
```
  query?: string;
  sort?: string;
  sortOrder?: string;
  aggregations?: string;
```
`format?` could stay as well because it's very likely something we'll want to filter by, OTOH `publicationDate` can def go as it's an article concept and is not relevant to events

jamieparkinson (approved):

api/src/controllers/events.ts
georgiaewhitney: Reflects this now - have kept `format?` if it'll be used eventually, but happy to remove if preferred

Issue #84, opened 2024-01-04
Events sorting


rcantin-w:
The default sorting option is relevance. Unsure yet if this should be offered as an option in the end result, but for now we want to ensure we offer a second option, which is probably `startDateTime`/`desc` (in Prismic), as it's what we currently use on `/events`.

georgiaewhitney:
sorted in #89

PR #86, opened 2024-01-04
Sort events by startDateTime

georgiaewhitney:
Sorts #84 

Start times added to index mapping and functionality to `sort` added into the events controller.

### DIFF
--- a/api/src/controllers/events.ts
+++ b/api/src/controllers/events.ts
@@ -7,6 +7,7 @@ import { Config } from "../../config";
 import { HttpError } from "./error";
 import { ResultList } from "../types/responses";
 import { resultListResponse } from "../helpers/responses";
+import { queryValidator, validateDate } from "./validation";
 
 type QueryParams = {
   query?: string;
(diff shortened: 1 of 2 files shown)


api/src/controllers/events.ts
```diff
@@ -2,11 +2,12 @@ import { errors as elasticErrors } from "@elastic/elasticsearch";
 import { RequestHandler } from "express";
 import asyncHandler from "express-async-handler";
 import { Clients, Displayable } from "../types";
-import { PaginationQueryParameters } from "./pagination";
+import { PaginationQueryParameters, paginationElasticBody } from "./pagination";
 import { Config } from "../../config";
 import { HttpError } from "./error";
 import { ResultList } from "../types/responses";
 import { resultListResponse } from "../helpers/responses";
+import { queryValidator, validateDate } from "./validation";
```
rcantin-w, line 10: I think `validateDate` is unused?

rcantin-w (approved):
Seems to make good sense to me!

api/src/controllers/events.ts
georgiaewhitney: Good spot! Removed

Issue #70, opened 2023-10-27
Create an Events unpublisher


rcantin-w:
**Phase 2**

rcantin-w:
merged to production

PR #87, opened 2024-01-04
create separate unpublishers for articles and events

agnesgaroux:
Issue https://github.com/wellcomecollection/content-api/issues/70

The `document-unpublish` event gives us no way to tell whether the `documents[]` are articles or eventDocuments and the unpublisher fails gracefully if the document id is not found in the index, so we're running all the ids through both the `articlesUnpublisher` and the `eventDocumentsUnpublisher` 

❓ I tried running `AWS_PROFILE=catalogue-developer PIPELINE_DATE="2023-03-24" yarn dev` and got
```
Error fetching secret 'prismic/content-unpublisher/secret' Secrets Manager can't find the specified secret.
```
I'm confused as to what this secret is/does, and I couldn't find it in secret manager. 
However I found `prismic/webhook/unpublish-secret` which is an env_var that the webhook-lambda compares to the secret in the webhook event before passing it on the unpublisher if they match
Should `prismic/content-unpublisher/secret` actually be `prismic/webhook/unpublish-secret`?
But then why should the unpublisher need it? the secret check happens before it's triggered

### DIFF
--- a/unpublisher/src/unpublisher.ts
+++ b/unpublisher/src/unpublisher.ts
@@ -1,0 +1,41 @@
+import {
+  Client as ElasticClient,
+  errors as elasticErrors,
+} from "@elastic/elasticsearch";
+import log from "@weco/content-common/services/logging";
+import { Result } from "@elastic/elasticsearch/lib/api/types";
+
+type Clients = {
+  elastic: ElasticClient;
+};
+
+export const createUnpublisher =
+  (index: string) =>
+  async (clients: Clients, unpublishedDocuments: string[]) => {
+    // There is no way to tell from the payload whether the update is for unpublished
+    // documents or published documents; we have to rely on having configured the
+    // webhook correctly.
+    // See https://prismic.io/docs/webhooks#a-document-is-unpublished
+    const deletedDocuments = Object.fromEntries<Result>(
+      await Promise.all(
+        unpublishedDocuments.map(async (id) => {
+          try {
+            const response = await clients.elastic.delete({
+              index,
+              id,
+            });
+            return [id, response.result] as const;
+          } catch (e) {
+            if (e instanceof elasticErrors.ResponseError) {
+              if (e.statusCode === 404) {
+                return [id, "not_found"] as const;
+              }
+            }
+            throw e;
+          }
+        })
+      )
+    );
+    log.info(`${Object.keys(deletedDocuments).length} deletions complete: `);
+    log.info(JSON.stringify(deletedDocuments));
+  };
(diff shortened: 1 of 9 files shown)


jamieparkinson:
I think that the secret fetching in the unpublisher must have been left over development code and should be deleted - most obvious sign that the unpublisher doesn't need it is that the variable `secret` is never used!

jamieparkinson (approved):

unpublisher/src/unpublisher.ts
```diff
@@ -0,0 +1,41 @@
...
+import { Result } from "@elastic/elasticsearch/lib/api/types";
+
+type Clients = {
+  elastic: ElasticClient;
+};
+
+export const createUnpublisher =
+  (index: string) =>
+  async (clients: Clients, unpublishedDocuments: string[]) => {
+    // There is no way to tell from the payload whether the update is for unpublished
```
kenoir, line 15: An aside: at present the prismic webhook lambda that is pushing events to eventbridge passes them through unchanged which perpetuates this publish/unpublish ambiguity coming from the prismic side. The webhook lambda(s) could do more work to transform the event into something more custom we consume internally to be clearer about the purpose of the event and omit unnecessary fields.

jamieparkinson: I don't necessarily disagree, but perhaps some useful context is that the webhook lambda was deliberately designed to pass the events through without a new model in order to separate the concerns of (a) a fixed endpoint for Prismic to call for a webhook and (b) pipeline/cluster-dependent handling of the events arriving via said webhook.

kenoir: I see. That makes sense, though i'd still favour having an internal event representation for the reasons above. You could have another step doing the transformation for internal services, [perhaps do that in EventBridge](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-transform-target-input.html) without another lambda (this is speculative, apart from a review of this PR).

jamieparkinson: Not sure if I have understood correctly but also worth noting that it would also be impossible for the webhook Lambda to know whether the event is a publish/unpublish - that ambiguity can't be resolved without additional API calls

kenoir: We'd have a webhook per event type (configured from Prismic) i'd guess and then the lambda for each surfaced by the function URL is configured to know it's for that type of event (not being able to infer it from the event itself but we tell it what it is when we set it up).

Not blocking for merge, but interesting to think about what we might do in the future.

kenoir: @jamieparkinson points out we already do this here: https://github.com/wellcomecollection/content-api/blob/main/infrastructure/webhook_stack/events.tf#L9-L11 ✨

PR #89, opened 2024-01-10
Fixes sorting asc/desc order

georgiaewhitney:
## What does this change?

- Adds the additional queries to the events mapping index, as well as ensuring dynamic is set to strict.
- `times: {startDateTime: Date[]` type added to ElasticsearchEventDocument
- The times property then added to the transformEventDocument in the eventDocument. 

## How to test

`events?sort=times.startDateTime&sortOrder=desc` and `events?sort=times.startDateTime&sortOrder=asc` should now display events in order accordingly.

### DIFF
--- a/pipeline/src/transformers/eventDocument.ts
+++ b/pipeline/src/transformers/eventDocument.ts
@@ -97,6 +97,8 @@ export const transformEventDocument = (
     id,
   } = document;
 
+  const documentTimes = document.data.times;
+
   const primaryImage = promo?.[0]?.primary;
   const image =
     primaryImage && isImageLink(primaryImage.image)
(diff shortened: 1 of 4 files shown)


pipeline/src/transformers/eventDocument.ts
```diff
@@ -126,6 +128,13 @@ export const transformEventDocument = (
       title: asTitle(title),
       caption: primaryImage?.caption && asText(primaryImage.caption),
       series: formatSeriesForQuery(document),
+      times: {
+        startDateTime: documentTimes
+          .map((time) =>
+            time.startDateTime ? new Date(time.startDateTime) : null
+          )
+          .filter((date): date is Date => date !== null),
```
jamieparkinson, line 136: General approach is to avoid `null`s in JS - this slightly over the top thought leadery blog post gives some good reasons https://www.split.io/blog/bias-against-null-javascript/;

Instead I would suggest first doing the transform from line 121 for `documentTimes`:
```typescript
const documentTimes = times.map(prismicTimestampToDate)
```
then you can do
```
times: documentTimes
```
on line 121 and then the following here
```suggestion
        startDateTime: documentTimes
          .map(time => time.startDateTime)
          .filter(isNotUndefined),
```

georgiaewhitney: Perfect - adjusted!

jamieparkinson (approved):
Lovely - these mapping changes will also set us up nicely to be able to do an initial events search

PR #97, opened 2024-01-25
Add default code owners

rcantin-w:
## What does this change?

I've added a default code owner so they automatically get added to PRs as reviewers. Let me know if it should be a different Team, but it makes sense to me as this repo is TS.

## How to test

I guess once it's in, and a PR gets created we could check it adds it automatically to reviewers?

Are PRs not going unnoticed anymore 👀 

I don't think there is any.


georgiaewhitney (approved):
Makes sense to me 👍

Issue #96, opened 2024-01-23
What do we do with events' isFullyBooked/onlineIsFullybooked?


agnesgaroux:
See https://github.com/wellcomecollection/content-api/issues/92
As discussed, we want the design of the event search result cards to closely match the one used in the event page
That includes the booking status ("fully booked") of the event which depends on Prismic's `isFullyBooked` and `onlineIsFullyBooked`. 
We're unsure how these are set and how they related/differ. Needs clarification from editorial before we add it into the event index

Issue #93, opened 2024-01-18
Events; scheduled events


rcantin-w:
As discussed; some events are considered "parent" events as they have a list of other events in their `Schedule` section:

<img width="1160" alt="Screenshot 2024-01-18 at 16 35 36" src="https://github.com/wellcomecollection/content-api/assets/110461050/95f8a848-f985-4fec-9230-e725502164f7">

These "scheduled/children" events are, in reality, sibling events as there is no hierarchy within the Event content type. They are created only to be listed [in the "main/parent" event](https://preview.wellcomecollection.org/events/ZQgdkREAAAbr6cRR):

<img width="682" alt="Screenshot 2024-01-18 at 16 36 49" src="https://github.com/wellcomecollection/content-api/assets/110461050/ea80f257-d736-4e44-af86-b50c962a54d8">

As less detailed events (the main information living within the "parent/main" event), they are not to be displayed in events listings. Editorial deals with this by adding the "`delist`" tag.
We also can't show them as they are not built in a way that makes them that useful/pleasing (no promo, no image, no description..), but we still need to be able to find them in a search. They contain important information such as date/time and access needs (interpretations). For example, a parent event would have no interpretation, but it might have a child event that is a BSL tour, and another one that is an Audio-described one. 

As discussed with @agnesgaroux , we think they could be added to the "parent/main" event in a new section, e.g. `childEvents` (or something less terrible), looking something like;

```
childEvents [
    {
      startDateTime: Date,
      endDateTime: Date,
      interpretations: string[]
    },
    {
      startDateTime: Date,
      endDateTime: Date,
      interpretations: string[]
    } 
]
```

As discussed with Jason, a "parent/main" event would then inherit its children's interpretation labels on the Front-end as well as on the BE, making it findable should a search be filtered by that interpretation.

rcantin-w:
Actually I wonder if `interpretations: string[]` should be passed fully instead of being just an array of titles, just so [it matches the type](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/wellcome/content/types/api.ts#L54) at the root level, which might be more expected and aligned.

agnesgaroux:
Just to make a distinction between API and index:
- we'd still index parent and children (scheduled) events separately 
- the indexed parent event has a `schedule`:
```
schedule: [
      {
        event: {
          id: 'ZQgi7hIAACQAN2QG',
          type: 'events',
[...]

PR #98, opened 2024-01-26
add booking info in times objects + childScheduledEvent at root level

agnesgaroux:
## What does this change?

As discussed, we're adding a property to indicate an event is a "child", ie. part of the schedule of a "parent" event, to be excluded from search results as the MVP stage https://github.com/wellcomecollection/content-api/issues/93
Also adding online/in-venue booking status/availability https://github.com/wellcomecollection/content-api/issues/96

## How to test

See updated test snapshots 

Once the changes are reflected in the index, FE will be able to display booking status on the event card in search
We can also exclude delisted events (without pictures) from search 

Feature behind a toggle so no risk

### DIFF
--- a/pipeline/src/types/transformed/index.ts
+++ b/pipeline/src/types/transformed/index.ts
@@ -69,7 +69,7 @@ export type ElasticsearchArticle = {
 
 export type ElasticsearchEventDocument = {
   id: string;
-  childScheduledEvent?: boolean;
+  isChildScheduledEvent?: boolean;
   display: EventDocument;
   query: {
     linkedIdentifiers: string[];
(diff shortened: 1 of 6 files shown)


pipeline/src/types/transformed/index.ts
```diff
@@ -69,6 +69,7 @@ export type ElasticsearchArticle = {
 
 export type ElasticsearchEventDocument = {
   id: string;
+  childScheduledEvent?: boolean;
```
rcantin-w, line 72: Could we rename to `isChildScheduledEvent` as we try to do that for booleans (`has`, `is`, type prefixes)

rcantin-w (approved):
Apart from the one comment I think this is good to go for `fullyBooked` stuff!

Issue #109, opened 2024-02-27
Add isAvailableOnline to content API


agnesgaroux:
`isAvailableOnline` has been added to the event index
https://github.com/wellcomecollection/content-api/issues/107

The API needs to be updated to accept a new query param `isAvailableOnline`

Add an eventFilter like 
```
isAvailableOnline: (): QueryDslQueryContainer => ({
    term: {
      "filter.isAvailableOnline": true,
    },
  }),
```
ie. if the query param is present, its value is `true`

Add an event aggregation like 
```
isAvailableOnline: {
    terms: {
      size: 2,
      field: "aggregatableValues.isAvailableOnline",
    },
  },
```

rcantin-w:
https://wellcomecollection.org/search/events?isAvailableOnline=true works nicely in prod

Issue #107, opened 2024-02-22
Add isAvailableOnline to event index


agnesgaroux:
Some "catch up" events have recordings, this is indicated by the the `availableOnline` on the Prismic document

The graph query already include the property
Update the transformer and associated types to include `isAvailableOnline` as a boolean 
Add to display, filters and aggregations 
Reindex with `reindex-events` from the `/pipeline` directory

PR #110, opened 2024-03-04
add isAvailableOnline to display, filter and aggValues for eventDocument

agnesgaroux:
## What does this change?
https://github.com/wellcomecollection/content-api/issues/107
Adds `isAvailableOnline` to eventDocuments so that we can display the value on the event search result cards, and filter and aggregate based on it
⚠️ Reindex not performed yet

## How to test

See test snapshot

Together with https://github.com/wellcomecollection/content-api/issues/109 we'll be able to search for/filter events that were recorded and can be watched online

Behind a toggle so minimal risk, can delete index and recreate it if something goes wrong

### DIFF
--- a/pipeline/src/transformers/eventDocument.ts
+++ b/pipeline/src/transformers/eventDocument.ts
@@ -121,7 +121,7 @@ export const transformEventDocument = (
   document: EventPrismicDocument
 ): ElasticsearchEventDocument => {
   const {
-    data: { title, promo, times, isOnline },
+    data: { title, promo, times, isOnline, availableOnline },
     id,
     tags,
   } = document;
@@ -156,6 +156,7 @@ export const transformEventDocument = (
       interpretations,
       audiences,
       series: transformSeries(document),
+      isAvailableOnline: availableOnline,
     },
     query: {
       linkedIdentifiers: linkedDocumentIdentifiers(document),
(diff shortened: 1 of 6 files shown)


rcantin-w (approved):
👀

agnesgaroux:
isAvailableOnline filter is 
```
isAvailableOnline: true || false
```
AggregatableValue is 
```
"isAvailableOnline": """{"isAvailableOnline":true}"""
```
This diverged from the common pattern of 
`"""{"type":"EventInterpretation","id":"Wn3STCoAACgAIedR","label":"Hearing loop"}"""` type/id/label
but is consistent with the structure of the `isAvailableOnline` property in the display object

pipeline/src/transformers/eventDocument.ts
```diff
@@ -182,6 +184,9 @@ export const transformEventDocument = (
         type: "EventLocation",
         isOnline: !!isOnline,
       }),
+      isAvailableOnline: JSON.stringify({
```
agnesgaroux, line 187: isAvailableOnline filter is 
```
isAvailableOnline: true || false
```
AggregatableValue is 
```
"isAvailableOnline": """{"isAvailableOnline":true}"""
```
This diverges from the common pattern of 
`"""{"type":"EventInterpretation","id":"Wn3STCoAACgAIedR","label":"Hearing loop"}"""` type/id/label
but is consistent with the structure of the `isAvailableOnline` property in the display object

PR #114, opened 2024-03-07
model prismic and transformed venue opening times

agnesgaroux:
In the Prismic type `modifiedDayOpeningTimes.type` is not necessary now that I think of it 

In the transformed type `id` could be the Prismic document id or we can created our own

We might also add a simple graph query if we don't want to rely on the hardcoded Prismic document id for library venue

### DIFF
--- a/pipeline/src/types/transformed/venue.ts
+++ b/pipeline/src/types/transformed/venue.ts
@@ -1,0 +1,26 @@
+export type Venue = {
+  type: "Venue";
+  id: string;
+  title: string;
+  regularOpeningDays: {
+    dayOfWeek: DayOfWeek;
+    opens: string;
+    closes: string;
+    isClosed?: boolean;
+  }[];
+  exceptionalOpeningDays: {
+    overrideDate: Date;
+    opens: string;
+    closes: string;
+    isClosed?: boolean;
+  }[];
+};
+
+type DayOfWeek =
+  | "monday"
+  | "tuesday"
+  | "wesdnesday"
+  | "thursday"
+  | "friday"
+  | "saturday"
+  | "sunday";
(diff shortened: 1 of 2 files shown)


pipeline/src/types/transformed/venue.ts
```diff
@@ -0,0 +1,26 @@
...
+    opens: string;
+    closes: string;
+    isClosed?: boolean;
+  }[];
+  exceptionalOpeningDays: {
+    overrideDate: Date;
+    opens: string;
+    closes: string;
+    isClosed?: boolean;
+  }[];
```
jamieparkinson, line 16: I wonder if the concept of "regular" vs "exceptional" opening days is something that we could hide - I understand why it's done this way in Prismic, but I don't know that it's helpful for API consumers (including ourselves!). The question that a client wants answered is just "when is the venue open?", and ideally they shouldn't have to work out the answer to that themselves by understanding regular and exceptional opening times.

I'm thinking that we could use this consideration for quite a nice separation of concerns with the date logic: the venue API (this one) tells you a list of opening dates, and the items API reuses that, potentially filtering it. After all, the venue opening times don't belong to the catalogue metadata, so it makes sense to reflect that in our API design.

Maybe in service of this we could try a simple model like this that returns the next 7 days (for example) by default
 ```suggestion
  openingDates: {
    open: Date,
    close: Date
  }[]
```

with flexibility enabled by a `openingDates.open.from` and a `openingDates.open.to` filter to request different ranges. Then all of the calculation of exceptional vs regular days would be completely hidden (and written in Typescript, which I suspect will be easier than doing this in Scala...). What do you think?

kenoir: @agnesgaroux is implementing so I defer to her, but this seems like a very good idea to me.

PR #115, opened 2024-03-13
Venue pipeline

agnesgaroux:
## What does this change?

This creates a pipeline to ingest Prismic `venue` documents into a new `venues` index
https://github.com/wellcomecollection/catalogue-api/issues/757

## How to test

Indexing can be performed without errors
Once indexing is successful, check documents in elastic console

The documents are successfully index in a new `venues` index

We can delete the index and start again if something goes wrong


pipeline/src/helpers/prismic.ts
```diff
@@ -17,8 +17,9 @@ const PRISMIC_MAX_PAGE_SIZE = 100;
 
 type GetPrismicDocumentsParams = {
   publicationWindow: TimeWindow;
-  graphQuery?: string;
+  graphQuery: string;
   after?: string;
+  documentType?: string;
```
kenoir, line 22: Can `documentType` ever be undefined?

kenoir:
If i'm reading this correctly we are reading `RegularOpeningDay` and `ExceptionalClosedDay` from Prismic and then calculating `nextOpeningDates` which is stored in ES. 

I think the `nextOpeningDates` calculation could be moved to the API after storing opening days and exceptions in ES as the current method here relies on regular extraction to generate upcoming dates, and i'm not sure this will function correctly if we are only picking up published changes.

pipeline/src/helpers/prismic.ts
agnesgaroux: It could, when this prismic helper was used by the 2 different ETL pipelines. I belatedly came to the realisation that the existing pipeline could do the job so that's gone now 🙃

agnesgaroux:
> If i'm reading this correctly we are reading `RegularOpeningDay` and `ExceptionalClosedDay` from Prismic and then calculating `nextOpeningDates` which is stored in ES.
> 
> I think the `nextOpeningDates` calculation could be moved to the API after storing opening days and exceptions in ES as the current method here relies on regular extraction to generate upcoming dates, and i'm not sure this will function correctly if we are only picking up published changes.

of course 🤦‍♀️

pipeline/src/indices/venues.ts
```diff
@@ -0,0 +1,34 @@
...
+      properties: {
+        id: {
+          type: "keyword",
+        },
+        title: {
+          type: "keyword",
+        },
+      },
+    },
+    nextOpeningDates: {
```
kenoir, line 23: Does this mapping need updating to reflect the change to calculate this value when serving from the api rather than at the transform? I think this needs to reflect the types of  `regularOpeningDays` and `exceptionalClosedDays`.

agnesgaroux: didn't use keyword type as we won't use the `display` bit for searching document, only the `filter` part

kenoir (approved):
👌

PR #120, opened 2024-03-19
Remove isOnline and create locations

rcantin-w:
## What does this change?

Removes our original idea of an `isOnline` boolean filter and creates a `locations` object that will serve both the needs of the FE display and the BE filtering/aggregations.

## How to test

Explore ElasticCloud as it's been reindexed just now, view the new objects.

We can happily work on the API and the FE parts without issue

Everything behind a toggle, not risky.


agnesgaroux:
Looking gooood

agnesgaroux (approved):

PR #122, opened 2024-03-21
create endpoint to get venues

agnesgaroux:
## What does this change?

https://github.com/wellcomecollection/content-api/issues/116
Adds `venues` to the content-api
can be queried by `id` or `title` eg. `/venues?title=library`

## How to test

Run locally and get `http://localhost:3000/venues?title=library`
Response should include regular opening dates, exceptional closures and the upcoming open days 

Negligible, the API is not yet used by any of our services


jamieparkinson (commented):
Sorry, I've been slightly out of the loop with this work and have some questions re this and https://github.com/wellcomecollection/content-api/pull/115/files. I think most of them probably just conversations I've missed!

- How come we needed to move the date logic from the pipeline to the API?
- We've chosen to still expose regular opening days and exceptional closure dates in the Venue display model, how come we no longer want to hide them?
- It looks like we're modifying the display model (adding `NextOpeningDate`) in the API, can we refactor so that the types are more explicit? I think for maintainability and consistency we should aim to have the contents of the `display` field in the pipeline of the same _type_ as the actual API document model even if we can't populate them yet. Everywhere else, we treat the `display` field as completely opaque JSON: if we need to use data from it, we duplicate that data in another field (eg for filters). Maybe we need to add a `data` field to hold the stuff to construct opening dates?
- The date logic is complex enough that it needs tests and comments

kenoir:
> How come we needed to move the date logic from the pipeline to the API?

Hopefully I can help explain this: we made the decision to do this as (a) upcoming dates won't be added correctly as triggering an update relies on changes happening in Prismic and (b) it makes sense to move the calculation to the content-api side in this case, because the transform on the source data is not dependent on changes to the source data (but instead the passage of time).

> The date logic is complex enough that it needs tests and comments

I agree that while there are tests for the functions in `util` further comments and controller tests would be very handy.

agnesgaroux:
> We've chosen to still expose regular opening days and exceptional closure dates in the Venue display model, how come we no longer want to hide them?
Thinking longer term (when all Prismic data would be proxied via content-api) weco.org will want the "raw" data for the `https://wellcomecollection.org/visit-us` and opening times pages. Since it was fairly straightforward to add it now, I did

As Robert pointed out, we have to generate the `NextOpeningDate[ ]` in the API because we don't write to the index unless the Prismic data changes, whereas the `NextOpeningDate[ ]` needs to be up-to-date at the time it's requested

agnesgaroux:
> we should aim to have the contents of the display field in the pipeline of the same type as the actual API document model even if we can't populate them yet

I'll have a think about this one

jamieparkinson:
Cool all helpful, thank you - has clarified that I think my main comment is that `display` should ideally be used to `display` only rather than for deriving data etc!

jamieparkinson (commented):
Comments are excellent - I can't see the tests for `getNextOpeningDates` etc in this PR, do they need to be moved from the `pipeline` app or something?

Still think that using `display` as something interpretable (rather than opaque JSON) is risky: for filters and aggregations we duplicate the data and I think we should do the same here

agnesgaroux:
Test for `getNextOpeningDates` are in api/test/utils.test.ts

jamieparkinson (approved):
Fab 🎉

PR #126, opened 2024-05-07
nextOpeningDates start at and include today

agnesgaroux:
## What does this change?

Following the discovery of a bug in the itemsAPI, we realised there was a mismatch between the content-api's venues response, and the way the itemsAPI was using it, ie. content-api starting the list of openingTimes at today+1 while itemsAPI assuming it started at "today" 
https://wellcome.slack.com/archives/C02ANCYL90E/p1714748440989459
It was decided that the content-api should respond with a list of openingTimes that start with "today" if today is an open day

## How to test

Running the api locally, hit `http://localhost:3000/venues?title=library` and check the `nextOpeningDates` start at today if today is an open day

ItemsAPI is the only client of this endpoint. There will be checks there and in weco.org before this change is used by live systems

### DIFF
--- a/api/src/controllers/utils.ts
+++ b/api/src/controllers/utils.ts
@@ -13,10 +13,7 @@ export function getNextOpeningDates(
   // create a dateList of 21 consecutive days starting tomorrow
   // 21 is arbitrary, we just need to have at least 12 once all the closed days have been removed
   const dateNow = new Date();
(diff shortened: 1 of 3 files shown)


kenoir (approved):
👍  Bit more context in the PR for future GitHub divers would be useful.

PR #127, opened 2024-05-07
Fix graph query for articles with slice machine

agnesgaroux:
## What does this change?

Following the slice machine changes reflected in the Prismic API, we have to change the graphQuery used to fetch articles from said Prismic API.

## How to test

The graph query is valid: running `yarn update-prismic-snapshots` doesn't error
The pipeline tests pass using the updated prismic snapshots

No more errors in the alert channel


pipeline/src/transformers/article.ts
```diff
@@ -109,7 +109,10 @@ export const transformArticle = (
 
   const queryBody = data.body
     ?.map((slice) => {
-      if (["text", "quoteV2", "standfirst"].includes(slice.slice_type)) {
+      if (
+        ["text", "quoteV2", "quote", "standfirst"].includes(slice.slice_type)
+      ) {
+        // quoteV2 can be removed once the slice Machine migration has completed
```
rcantin-w, line 115: 👍

rcantin-w:
[Slack chats happened here ](https://wellcome.slack.com/archives/CUA669WHH/p1715087740894799)

PR #132, opened 2024-07-30
get day name in London timezone + test around midnight

agnesgaroux:
## What does this change?
https://github.com/wellcomecollection/content-api/issues/133
[Slack thread](https://wellcome.slack.com/archives/CUA669WHH/p1722266021986019)
LE&E brought to our attention that a library user was able to request an item for a Sunday.
-> On Friday between 11pm and midnight during BST, a timezone loophole caused the Sunday to be mistaken for a Monday, thus to not be removed from the available date list.
This is fixed by explicitly setting the timezone as `["Europe/London"]` when we extract the day of the week from the timestamp

## How to test

The tests have been duplicated to cover the same scenarios both in UTC and BST times of the year.
An additional test case specifically covers the "around midnight on a Friday during BST" scenario

When placing an item request around midnight on a Friday during BST, the 1st date in the datePicker dropdown is not Sunday but Monday

LE&E have shown that they can provide a layer of defence by catching erroneous requests and alerting us to unexpected behaviour

### DIFF
--- a/api/test/utils.test.ts
+++ b/api/test/utils.test.ts
@@ -66,6 +66,13 @@ const exceptionalClosedDays = [
     startDateTime: "00:00",
     endDateTime: "00:00",
   },
+  {
+    // this is during BST so 2024-07-28T23:00:00.000Z is 2024-07-29T00:00:00.BST
+    overrideDate: "2024-07-28T23:00:00.000Z",
+    type: "Summer Bank Holiday",
+    startDateTime: "00:00",
+    endDateTime: "00:00",
+  },
 ];
 
 const mockDateNow = (dateToMock: string) => {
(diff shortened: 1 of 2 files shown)


rcantin-w (approved):
The fix makes sense to me! 


(Non blocker as tests are happy): I just wonder about how we do it for [`getDateWithoutTime`](https://github.com/wellcomecollection/content-api/pull/132/files#diff-905804f881e0a39d0f7645a6cfbda2a906c72ab83a4a85de0dd65915ba05539eR60) and if we should align. Or should the zone be set at a higher level if we need it to be London time throughout the process (I have less context than you, so very much a question)?

agnesgaroux:
@rcantin-w the `toLocaleString` that `getDateWithoutTime` uses is from `luxon` while `getDayName` uses vanilla javascript so the syntax is slightly different but `getDateWithoutTime` does in fact set the timezone to `["Europe/London"]` 
```
function getDateWithoutTime(date: Date): string {
  return DateTime.fromJSDate(date).setZone("Europe/London").toLocaleString();
}
```

api/test/utils.test.ts
```diff
@@ -66,78 +66,189 @@ const exceptionalClosedDays = [
...
 ];
 
 const mockDateNow = (dateToMock: string) => {
   jest.useFakeTimers().setSystemTime(new Date(dateToMock));
 };
 
 describe("getNextOpeningDates", () => {
-  it("start the dateList from today", () => {
-    mockDateNow("2024-03-12T08:00:00.000Z");
+  describe("when London is in UTC", () => {
```
kenoir, line 83: Do we need to mock the timezone to UTC to simulate the deployed lambda environment? 

See the [AWS docs for the unreserved lambda env var `TZ`](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html).

Something like https://www.npmjs.com/package/timezone-mock might help?

agnesgaroux: It somewhat baffles me, but I don't think we do.
The problem was not **where** the code runs, because wherever you call `new Date()` it returns a UTC timestamp which is the same everywhere for a specific point in time
The issue was trying to infer the **day of the week** from that UTC timestamp. Depending on where you are, the same exact point in time can be Saturday or Sunday, and that's addressed by
```
{ weekday: "long", timeZone: "Europe/London" }
```
ie. give me the day of the week for the point in time, **in London**

I actually installed `timezone-mock` and added 
```
beforeAll(() => {
  timezone_mock.register("Australia/Adelaide")
});

afterAll(() => {
  timezone_mock.unregister()
});
```
to the test file. Tests pass whether I set the timezone to UTC, London or even Adelaide

kenoir (approved):

PR #134, opened 2024-08-07
longer list of nextOpeningDates to account for deepstore leadtime

agnesgaroux:
## What does this change?

The `/venues` endpoint returns a longer list of `nextOpeningDays` so that we still have enough library opening days once we remove the 10 days of deepstore delivery time. We want to avoid what's in the screenshot below 👇 

Screenshot 2024-08-06 at 13 34 22

## How to test

https://api.wellcomecollection.org/content/v0/venues?title=library -> `nextOpeningDates`.length should be at least 22

Deploy the change to staging, the dropdown for deepstore/offsite items (see screenshot) should have the usual 12 days to choose from 

Given that the starting `dateList` is now 35 days long, even when we remove 5 Sundays we'd have to have 8 exceptional closed days (excluding Sundays) over the period for the list to become shorter than 12 days


api/test/venues.test.ts
```diff
@@ -72,26 +65,26 @@ const regularOpeningDays = [
...
+    overrideDate: "2024-04-08T00:00:00.000Z",
+    type: "Free bank holiday weekend!",
     startDateTime: "00:00",
     endDateTime: "00:00",
   },
   {
-    overrideDate: "2024-03-31T23:00:00.000Z",
-    type: "Easter",
+    overrideDate: "2024-04-30T23:00:00.000Z",
+    type: "Fête du travail",
```
rcantin-w, line 87: 🇫🇷

api/src/controllers/utils.ts
```diff
@@ -13,7 +13,7 @@ export function getNextOpeningDates(
   // create a dateList of 21 consecutive days starting tomorrow
   // 21 is arbitrary, we just need to have at least 12 once all the closed days have been removed
```
rcantin-w, line 14: Should the comment be updated?

api/test/venues.test.ts
```diff
@@ -16,13 +16,6 @@ describe("GET /venues", () => {
       nextOpeningDates: expectedNextOpeningDates,
     });
   });
-
-  it("returns a 404 if no document for the given ID exists", async () => {
-    const api = mockedApi([]);
-
-    const response = await api.get(`/articles/123`);
-    expect(response.statusCode).toBe(404);
-  });
```
rcantin-w, line 25: Was this removed on purpose?

agnesgaroux: Yes, there is no `/venues/{id}` endpoint. 
Bad copy/paste job when the test was originally added 😅

api/src/controllers/utils.ts
agnesgaroux: ✅

api/test/venues.test.ts
rcantin-w: Could I get more context on changing the dates/how it affects other tests and why the type changed as well?

rcantin-w (approved):

api/test/venues.test.ts
agnesgaroux: I only changed the dates to give us extra reassurance that the `exceptionalClosedDays` are in fact removed from the API response `nextOpeningDays` list. The previous `exceptionalClosedDays` didn't fall within the timeframe we were requesting
The type comes from Prismic originally and doesn't affect the result

Issue #111, opened 2024-03-04
Add linting to repo 


rcantin-w:
The current linting setup doesn't really seem to work. It'd be great to review it and make it impossible to commit code that doesn't match our standards (no unused imports, similar tabbing style, etc.)

The Typescript GitHub action we've added to .org should also be added to this repo.

rcantin-w:
As discussed with @kenoir this week; we might want to look into a solution that ensure we have the same linting rules on all TS projects (so wc.org and content-api at least).

rcantin-w:
Merged in!

PR #143, opened 2024-09-20
Add same linting and prettier config as wc.org

rcantin-w:
## What does this change?

Relates to #111 

1. Added linting! We didn't have any and we had things like unused imports. I went ahead and added the one we recently added to wc.org ([here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11170) then [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11178)).
  I removed the ones that had to do with React, but I think that's mostly the only difference.

2. I also upgraded to the lastest version of `prettier` and added the same config as wc.org, which affected a LOT of double quotes.

3. I aimed to split this in useful commits, so I'd recommend reviewing using those instead!
  

(note: for the `any` types I chose to not touch most of them as we did in .org, but those that I have tried to address I've left a comment on)

## How to test

Run things locally, do scripts work ok? (careful with pipeline scripts as they affect our one instance).
You can try `yarn lint` at the root too.

Readable and aligned code facilitating onboarding and further coding.

It could break something 🤷‍♀️


api/src/controllers/events.ts
```diff
@@ -71,11 +73,10 @@ const isAvailableOnlineValidator = queryValidator({
...
 
-  params["location"]
-    ? locationsValidator({
-        location: params["location"],
-      })
-    : undefined;
+  if (params.location)
+    locationsValidator({
+      location: params.location,
+    });
```
rcantin-w, line 79: This throws an error if not valid

api/src/controllers/venue.ts
```diff
@@ -34,39 +36,33 @@ const venuesController = (clients: Clients, config: Config): EventsHandler => {
 
     const filters = pickFiltersFromQuery(["title", "id"], params, venuesFilter);
 
-    try {
```
rcantin-w, line 37: This flagged try/catch as redundant, since it only throws the error it would throw anyway

pipeline/src/helpers/type-guards.ts
```diff
@@ -1,11 +1,15 @@
...
+  DataInterface,
+  PrismicImage,
+} from "@weco/content-pipeline/src/types/prismic";
 
 export function isNotUndefined<T>(val: T | undefined): val is T {
   return typeof val !== "undefined";
 }
 
-function isString(v: any): v is string {
+function isString(v: unknown): v is string {
```
rcantin-w, line 12: Again, I _believe_ this is the valid type in this case, feel free to contradict though!

pipeline/src/transformers/article.ts
```diff
@@ -82,9 +84,9 @@ function transformLabelType(
 }
 
 export const transformArticle = (
-  document: ArticlePrismicDocument
+  document: ArticlePrismicDocument,
 ): ElasticsearchArticle => {
-  const { data, id, first_publication_date } = document;
+  const { data, id, first_publication_date: firstPublicationDate } = document;
```
rcantin-w, line 89: Camel case gets flagged, but since we need this (it's how it is in Prismic), naming it something else calms the alert down.

pipeline/src/transformers/venue.ts
```diff
@@ -98,7 +99,7 @@ export const transformVenue = (
         title,
         title
           .toLowerCase()
-          .replace(new RegExp(/[èéêë]/g), "e")
+          .replace(/[èéêë]/g, "e")
```
rcantin-w, line 102: Warned as redundant

pipeline/test/event.test.ts
```diff
@@ -11,8 +11,8 @@ describe("time window events", () => {
       const result = toBoundedWindow(event);
 
       expect(result.end).toBeDefined();
-      expect(result.end?.getTime()! - result.start?.getTime()!).toBe(
-        60 * 60 * 1000
+      expect(result.end!.getTime() - result.start!.getTime()).toBe(
+        60 * 60 * 1000,
```
rcantin-w, line 15: My favourite telling off:
> Optional chain expressions can return undefined by design - using a non-null assertion is unsafe and wrong.

Here I assumed we knew end and start would always have a value

agnesgaroux:
Remove prettier formatting [here](https://github.com/wellcomecollection/content-api/blob/e0125376130465cbd1dd7820a4de8d7161d3a8da/.buildkite/scripts/autoformat.sh#L8)

pipeline/src/helpers/type-guards.ts
agnesgaroux: Sounds more accurate in this case 👍

pipeline/test/event.test.ts
agnesgaroux: > is unsafe and wrong.

There are a few times in my life when I wish I'd had this warning 😅

agnesgaroux: Just to clarify these are not Prismic events, but the event that trigger the lambda. 
This is testing that in the case where we have a `duration` and a `start` in the aforementioned event then the so-called "boundedWindow" does have an end and the time between start and end is as expected

pipeline/test/transformers/identifiers.test.ts
```diff
@@ -37,14 +37,14 @@ describe("linkedDocumentIdentifiers", () => {
     expect(identifiers).toIncludeSameMembers(["abc", "def"]);
   });
 
-  it("ignores linked documents without data", () => {
+  it("ignores linked documents without data", async () => {
```
agnesgaroux: Why do we need async here? `linkedDocumentIdentifiers` is synchronous

agnesgaroux: Ok, why `'expect' matchers must be awaited or returned`? The one above is not getting eslint's knickers in a twist

agnesgaroux: Fixed by replacing the playwright/recommended rules with jest/recommended

agnesgaroux (approved):

pipeline/test/event.test.ts
rcantin-w: Do you specify because you think the test could be clearer? (as in should we amend the description?)

PR #145, opened 2024-10-03
Add "ignorePattern" to eslint and add .vscode settings

rcantin-w:
## What does this change?

`yarn lint` wouldn't run well for me after having run many scripts that created output - it had too much to look at. In .org we use `.eslintignore`, but that's going to be deprecated once we switch to eslint 9, so I went with `ignorePatterns`. I've also copied the `.vscode/settings` from .org so we'd have the same set-up on both. I've also removed `.tsx` from files that should be linted as we don't have any in this repo.

## How to test

Run lint locally

Better alignment between projects

N/A


agnesgaroux:
Not a blocker: we could also remove `tsx` from the overrides/files lines 83 and 87

agnesgaroux (approved):

agnesgaroux:
> > Not a blocker: we could also remove tsx from the overrides/files in .eslintrc.js lines 83 and 87
> 
> Nice catch! Pushing now

The limited options of emojis to react to Github comments are really cramping my style

rcantin-w:
> > > Not a blocker: we could also remove tsx from the overrides/files in .eslintrc.js lines 83 and 87

> > Nice catch! Pushing now

> 🙌
> The limited options of emojis to react to Github comments are really cramping my style

😂

Issue #140, opened 2024-09-18
Add `uid` to Content API response


gestchild:
Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/11074

The content API response doesn't include the uid of the Prismic documents in the results

[We need this in order to construct links](https://github.com/wellcomecollection/wellcomecollection.org/pull/11190).

### To do:
- [x] `/pipeline`: Fetch UID from Prismic and ensure it gets added in ElasticSearch (#146) 
- [x] `/api`: Return UID in API response
- [x] Probably need to [render new test files](https://github.com/wellcomecollection/content-api/blob/main/pipeline/README.md#tests) and ensure the tests run successfully in both.

PR #146, opened 2024-10-07
Send uid to elastic

davidpmccormick:
For [#140](https://github.com/wellcomecollection/content-api/issues/140)

## What does this change?

It gets the `uid` from the Prismic document for `Article`s and `Events` and sends it along in the Elasticsearch transformers.

## How to test

- `cd pipeline`
- `yarn test`

Gets us closer to being able to use `uid`s instead of `id`s in our urls (relates to [11074](https://github.com/wellcomecollection/wellcomecollection.org/issues/11074))

I'm new here!


pipeline/src/transformers/eventDocument.ts
```diff
@@ -166,6 +167,7 @@ export const transformEventDocument = (
 
   return {
     id,
+    uid,
     ...(tags.includes('delist') && { isChildScheduledEvent: true }),
     display: {
```
rcantin-w, line 172: Should the UID be added here as well?

pipeline/src/types/transformed/index.ts
```diff
@@ -46,6 +46,7 @@ export type {
 
 export type ElasticsearchArticle = {
   id: string;
+  uid: string | null;
```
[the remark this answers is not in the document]
rcantin-w: Also, it shouldn't really happen but it has the potential too (e.g. we add the uid field to existing documents)

rcantin-w:
Left a few comments but wanted to say it otherwise looked good and nice to have the snapshots updated 👍 

This can be tested once in staging too, using our toggles and accessing the staging API, so we have that extra safety check before deploying to prod.

kenoir:
Do the index mappings also need updating to ensure we can index the new UID field properly?

pipeline/src/transformers/article.ts
```diff
@@ -87,6 +87,7 @@ export const transformArticle = (
   document: ArticlePrismicDocument
 ): ElasticsearchArticle => {
   const { data, id, first_publication_date: firstPublicationDate } = document;
+  const uid = document.uid || undefined;
```
agnesgaroux, line 90: Can it not be destructured like the other `document` attributes?

pipeline/src/transformers/eventDocument.ts
```diff
@@ -143,6 +143,7 @@ export const transformEventDocument = (
     id,
     tags,
   } = document;
+  const uid = document.uid || undefined;
```
agnesgaroux, line 146: Ditto

agnesgaroux:
LGTM! 👍 What's the rationale for having the uuid at the document's root level? Would it work to only have it in the `display` object?

agnesgaroux (approved):

Issue #155, opened 2024-11-15
Create graphQuery for each content type


rcantin-w:
We want to create a graphQuery that will only fetch what we need from each addressable content type (see [article's](https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/graph-queries/articles.ts)), so there's a piece of work to research and build it.

See https://github.com/wellcomecollection/content-api/issues/152 for what is needed for the `description` field

- [x] Books
- [x] Events
- [x] Exhibition texts
- [x] Exhibitions
- [x] Highlight tours
- [x] Pages
- [x] Projects
- [x] Seasons
- [x] Stories
- [x] Visual stories

PR #158, opened 2024-11-19
Add addressableQueries

davidpmccormick:
For #155 

## What does this change?
Adds each of the queries required for 'All' search individually

## How to test
`yarn graphQueryTest --type={type} --isAddressable=true` (ht @rcantin-w for the script)

Gets us closer to 'All' search

n/a

### DIFF
--- a/pipeline/src/graph-queries/addressables/articles.ts
+++ b/pipeline/src/graph-queries/addressables/articles.ts
@@ -1,0 +1,47 @@
+const query = `
+  articles {
+    title
+    format {
+      title
+    }
+    contributors {
+      role {
+        title
+      }
+      contributor {
+        ...on people {
+          name
+        }
+      }
+    }
+    promo {
+      ...on editorialImage {
+        non-repeat {
+          caption
+        }
+      }
+    }
+    body {
+      ...on text {
+        variation {
+          ...on default {
+            primary {
+              text
+            }
+          }
+        }
+      }
+      ...on standfirst {
+        variation {
+          ...on default {
+            primary {
+              text
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+export default query;
--- a/pipeline/src/graph-queries/addressables/articles.ts
+++ b/pipeline/src/graph-queries/addressables/articles.ts
@@ -5,9 +5,6 @@ const query = `
       title
     }
     contributors {
-      role {
-        title
-      }
       contributor {
         ...on people {
           name
--- a/pipeline/src/graph-queries/addressables/highlightTours.ts
+++ b/pipeline/src/graph-queries/addressables/highlightTours.ts
@@ -1,0 +1,19 @@
+const query = `
+  exhibition-highlight-tours {
+    title
+    intro_text
+    slices {
+      ...on guide_stop {
+        variation {
+          ...on default {
+            primary {
+              title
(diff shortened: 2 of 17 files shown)


rcantin-w (commented):
Should we add `webcomics`? (Although that might be a question; is it worth it since we want to move them over to being `articles` soon-ish, and they barely have queriable information...)

pipeline/src/scripts/testGraphQuery.ts
```diff
@@ -72,21 +90,77 @@ async function main() {
...
+        };
       case 'exhibition-text':
-        return { graphQuery: '', id: id || 'Zs8mohAAAB4AP4sc' };
+        return {
+          graphQuery: isAddressable
+            ? `{
+            ${addressablesExhibitionTextsQuery}
+          }`
+            : '',
+          id: id || 'Zs8mohAAAB4AP4sc',
```
rcantin-w, line 135: This is not for this ticket and is a very general ponder; should we move it the other way around since most of these are Addressables? The flag being `isSpecific` or something instead.

pipeline/src/graph-queries/addressables/visualStories.ts
```diff
@@ -0,0 +1,14 @@
...
+    title
+    promo {
+      ...on editorialImage {
+        non-repeat {
+          caption
+        }
+      }
+    }
+  }
+`;
```
rcantin-w, line 12: [In the table](https://www.notion.so/wellcometrust/All-search-definition-acceptance-criteria-1366687658a180f28254d4429bf764ce) we say we could use standfirst, but I know in your findings you opted for promo text over it. Should it still be used for query?

pipeline/src/graph-queries/addressables/highlightTours.ts
```diff
@@ -0,0 +1,19 @@
...
+    intro_text
+    slices {
+      ...on guide_stop {
+        variation {
+          ...on default {
+            primary {
+              title
+              transcript
+            }
+          }
```
rcantin-w, line 13: Nice thinking here!

pipeline/src/scripts/testGraphQuery.ts
rcantin-w: If you think it's a good idea I'm happy to do it as part of https://github.com/wellcomecollection/content-api/pull/161

pipeline/src/graph-queries/addressables/articles.ts
```diff
@@ -0,0 +1,47 @@
+const query = `
+  articles {
+    title
+    format {
+      title
+    }
+    contributors {
+      role {
+        title
+      }
```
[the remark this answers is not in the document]
davidpmccormick: yeah I wasn't sure about this and erred on the side of caution. Shall I get rid of it?

pipeline/src/graph-queries/addressables/visualStories.ts
davidpmccormick: I think not. I'll update the table.

davidpmccormick:
I _think_ on balance I'd probably not include webcomics because like you say there is so little of queryable relevance, _and_ we're planning to migrate them

rcantin-w (approved):

Issue #156, opened 2024-11-15
Build transformers for each content types


rcantin-w:
Build transformers as per transformed documents results in the ["All" search RFC](https://github.com/wellcomecollection/docs/blob/all-search-rfc/rfcs/064-content-api-all-search/README.md)

- [x] Books https://github.com/wellcomecollection/content-api/pull/163
- [x] Events https://github.com/wellcomecollection/content-api/pull/169
- [x] Exhibition texts #171 
- [x] Exhibitions #172 
- [x] Exhibition highlight tours #173
- [x] Pages #176 
- [x] Projects
- [x] Seasons
- [x] Stories https://github.com/wellcomecollection/content-api/pull/170/
- [x] Visual stories https://github.com/wellcomecollection/content-api/pull/164
- [x] Parent transformer file that sends each document through its relevant transformer https://github.com/wellcomecollection/content-api/pull/162
- [x] Filter out `delist`ed (scheduled) events before they reach the transformer #179 
- [x] Allow for arrays to be returned from transformer #181
- [x] create new pipeline?

Issue #153, opened 2024-11-14
Index addressable content types


rcantin-w:
Index addressable content types in a new Elasticsearch index as per the ["All" search RFC](https://github.com/wellcomecollection/docs/blob/all-search-rfc/rfcs/062-content-api-all-search/README.md).

This ticket will also ensure the correct mapping is defined.

### Ready when these are done:
- [x] #155 
- [x] #156

rcantin-w:
I'll do the prep for this although it needs #156, it'll just be ready for when we have transformers. https://github.com/wellcomecollection/content-api/pull/162

PR #162, opened 2024-11-21
Update types and add helpers to index addressables

rcantin-w:
## What does this change?
#153  and #156

Adds necessary types and helpers to index Addressables.
I've also added the "parent" transformer file that sends documents to their relevant transformer.

## How to test

We won't index right away I think, which is why some code is commented out, so no need to test the indexing as part of this PR. As long as TS and tests are happy, so are we for now I think?

Eases further work

This isn't used yet so shouldn't have risks.

### DIFF
--- a/pipeline/src/handler.ts
+++ b/pipeline/src/handler.ts
@@ -11,6 +11,7 @@ import {
 } from './graph-queries';
 // import { addressablesQuery } from './graph-queries/addressables';
 import { articles, events, venues } from './indices';
+// import { transformAddressable } from './transformers/addressables';
 import { transformArticle } from './transformers/article';
 import { transformEventDocument } from './transformers/eventDocument';
 import { transformVenue } from './transformers/venue';
@@ -31,7 +32,7 @@ import { Clients } from './types';
 //     'seasons',
 //     'visual-stories',
 //   ]),
-//   transformer: transformAddressables,
+//   transformer: transformAddressable,
 // });
 
 const loadArticles = createETLPipeline({
--- a/pipeline/src/handler.ts
+++ b/pipeline/src/handler.ts
@@ -9,11 +9,30 @@ import {
   webcomicsQuery,
   wrapQueries,
 } from './graph-queries';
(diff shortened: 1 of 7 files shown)


pipeline/src/handler.ts
```diff
@@ -9,12 +9,32 @@ import {
...
+//     'exhibition-texts',
+//     'exhibition-highlight-tours',
+//     'pages',
+//     'projects',
+//     'seasons',
+//     'visual-stories',
+//   ]),
+//   transformer: transformAddressable,
+// });
+
```
rcantin-w, line 37: As discussed; We could uncomment this, but then I'd say we only put in the content types who have transformers

pipeline/src/indices/index.ts
davidpmccormick, line 1: who indexes the indices?

davidpmccormick (approved):

pipeline/src/indices/index.ts
rcantin-w: Two ways:
1. Manually by running one of the `reindex` yarn commands
2. The scheduled reindex that runs every 15 mins.

They both use the ETL, and the indexing happens [here](https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/extractTransformLoad.ts#L93). They know what parameters to use for the ETL because of [the handlers](https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/handler.ts#L18-L37). Which one is used is decided by [createHandler](https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/handler.ts#L39-L54). 

1. The manual run passes either a specific content type or all of them, and fetches every one of those documents that Prismic holds.
2. The scheduled reindex runs "all", but only fetches the relevant documents (from those content types) that have been published in the last 15 mins.

Because my code [comments outs the createHandler for addressables](https://github.com/wellcomecollection/content-api/pull/162/files#diff-fed15b9adccfeb90791020c7b1bb85e8504997525b77663c5a01094532bd6bd7R65-R67), they shouldn't be indexed by the automated reindex.

Does that answer your question?

PR #163, opened 2024-11-21
Add addressable book transformer

davidpmccormick:
## What does this change?
First go at the first of the addressable transformers needed for #156

## How to test
`cd pipeline && yarn testTransformer --type book`

image

We're a step closer to being able to display addressable content in 'All search'

Can't think of any

- [ ] @LaurenFBaily to confirm we are happy with `title: subtitle` in the `display` block

### DIFF
--- a/pipeline/src/scripts/testTransformer.ts
+++ b/pipeline/src/scripts/testTransformer.ts
@@ -116,14 +116,12 @@ async function main() {
 
       case 'book': {
         const doc = await client.getByID(id || 'WwVK3CAAAHm5Exxr', {
-          graphQuery: isDetailed
-            ? ''
-            : `{
+          graphQuery: `{
             ${addressablesBooksQuery}
           }`,
         });
 
-        transformerName = isDetailed ? '' : 'transformAddressableBook';
+        transformerName = 'transformAddressableBook';
         return transformAddressableBook(doc as BookPrismicDocument);
       }
 
--- a/pipeline/src/scripts/testTransformer.ts
+++ b/pipeline/src/scripts/testTransformer.ts
@@ -9,6 +9,7 @@ import {
 } from './../graph-queries';
 import { addressablesBooksQuery } from './../graph-queries/addressables';
 import { createPrismicClient } from './../services/prismic';
+import { transformAddressableBook } from './../transformers/addressables/book';
 import { transformArticle } from './../transformers/article';
 import { transformEventDocument } from './../transformers/eventDocument';
 import { transformVenue } from './../transformers/venue';
@@ -16,6 +17,7 @@ import {
   ArticlePrismicDocument,
   EventPrismicDocument,
 } from './../types/prismic';
+import { BookPrismicDocument } from './../types/prismic/books';
 import { VenuePrismicDocument } from './../types/prismic/venues';
 
 const { type, isDetailed, id } = yargs(process.argv.slice(2))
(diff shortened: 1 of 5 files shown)


pipeline/src/types/prismic/addressables/books.ts
davidpmccormick, line 1: I'm not sure if these types (and the future addressable ones) should live in their own `/addressables` dir, or if that's overkill (I think it might be)?

rcantin-w: I think it makes sense, seeing as we'll do that everywhere else (grouping them together).

pipeline/src/transformers/addressables/book.ts
```diff
@@ -0,0 +1,53 @@
...
+  const title = asTitle(data.title);
+  const subtitle = data.subtitle ? asText(data.subtitle) : undefined;
+  const titleSubtitle = `${title}${subtitle ? `: ${subtitle}` : ''}`;
+  const contributors = (data.contributors ?? [])
+    .map(c => {
+      return isFilledLinkToDocumentWithData(c.contributor)
+        ? asText(c.contributor?.data.name)
+        : undefined;
+    })
+    .filter(isNotUndefined);
```
rcantin-w, line 25: genuine don't know; could there be more contributors than just the author? Should we only get Authors? (it might not be a real issue)

pipeline/src/types/transformed/index.ts
```diff
@@ -96,3 +96,23 @@ export type ElasticsearchEventDocument = {
...
+    title: string;
+    description?: string;
+    contributors: string[];
+  };
+  query: {
+    type: 'Book';
+    title: string;
+    description?: string;
+    contributors: string[];
+    body?: string[] | string;
```
rcantin-w, line 116: I think the transformer makes it `string | undefined`

rcantin-w (approved):
Few comments, but looking good !

pipeline/src/transformers/addressables/book.ts
davidpmccormick: Apparently so https://wellcomecollection.org/books/brains

pipeline/src/types/prismic/addressables/books.ts
davidpmccormick: I opted just to move it into the `/prismic` dir instead, since the thing it references will be a full Prismic 'book' document

pipeline/src/scripts/testTransformer.ts
```diff
@@ -111,14 +114,18 @@ async function main() {
...
+            ${addressablesBooksQuery}
+          }`,
+        });
 
-      //   transformerName = 'transformVenue';
-      //   return transformVenue(doc as VenuePrismicDocument);
-      // }
+        transformerName = isDetailed ? '' : 'transformAddressableBook';
+        return transformAddressableBook(doc as BookPrismicDocument);
+      }
```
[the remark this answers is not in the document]
davidpmccormick: same for line `119` above then?

pipeline/src/transformers/addressables/book.ts
rcantin-w: Let's leave it like that and if we get comments we'll amend.

davidpmccormick: I think if there was e.g. an author and an illustrator listed in the contributors, we'd want to display both in the UI. I've made it a string and comma separated them for now

PR #169, opened 2024-11-26
Events addressable transformer

davidpmccormick:
## What does this change?
Adds a transformer for 'addressable' events, as part of the 'All' search work (#156)

## How to test
`cd pipeline && yarn testTransformer --type event --isDetailed false`
image

Gets us closer to being able to search all addressable content using our own api

can't think of any


pipeline/src/transformers/addressables/event.ts
```diff
@@ -0,0 +1,73 @@
...
+        end: asDate(t.endDateTime),
+      };
+    })
+    .filter(d => d.start && d.end)
+    .sort((a, b) => Number(a.start) - Number(b.start));
+
+  const firstStart = startEnd?.at(0)?.start;
+  const lastEnd = startEnd?.at(-1)?.end;
+
+  const times = {
```
davidpmccormick, line 33: We return the first possible start dateTime and the last possible end dateTime. We could have turned it into a string here, but we'd lose the ability to present it within `<time>` elements on the frontend, and this way we'll (hopefully) be able to reuse some of our pre-existing date range presentation logic without having to write any new date helper functions (which are already battle-tested in the front-end app)

pipeline/src/transformers/addressables/event.ts
```diff
@@ -0,0 +1,73 @@
...
+  const contributors = (data.contributors ?? [])
+    .map(c => {
+      return isFilledLinkToDocumentWithData(c.contributor)
+        ? asText(c.contributor?.data.name)
+        : undefined;
+    })
+    .filter(isNotUndefined)
+    .join(', ');
+
+  return tags.includes('delist')
```
davidpmccormick, line 52: This is speculative: rather than returning the transformed event with an [`isChildScheduledEvent`](https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/transformers/eventDocument.ts#L171) boolean attached, we just return nothing, because we (probably?) don't want these to be indexed at all

davidpmccormick: Opted against this approach and we'll bail out before these `delist`ed events reach this transformer stage

https://wellcome.slack.com/archives/C02ANCYL90E/p1732639950071859?thread_ts=1732638374.646629&cid=C02ANCYL90E

rcantin-w (commented):
Could you add it to `pipeline/src/transformers/addressables.ts` (I merged it in recently so it wouldn't have been there for the start of this) 🙏

pipeline/src/transformers/addressables/event.ts
```diff
@@ -0,0 +1,74 @@
...
+  const title = asTitle(data.title);
+  const uid = documentUid || undefined;
+  const contributors = (data.contributors ?? [])
+    .map(c => {
+      return isFilledLinkToDocumentWithData(c.contributor)
+        ? asText(c.contributor?.data.name)
+        : undefined;
+    })
+    .filter(isNotUndefined)
+    .join(', ');
```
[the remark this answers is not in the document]
rcantin-w: Thinking about it I think we could do a `transformGenericFields` [like we do in wc.org ](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/prismic/transformers/index.ts#L139). But we can do it at the end when we've got everything as a quick tidy. What do you think?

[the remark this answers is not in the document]
davidpmccormick: sounds good

rcantin-w (approved):
🎉
tests ran well locally, as did `testTransformer` 👍

PR #170, opened 2024-11-26
Addressable stories transformer and test changes

rcantin-w:
## What does this change?

#156 

- Adds Addressable Article transformer (and add that to `testTransformer`)
- Based on [this Slack conversation](https://wellcome.slack.com/archives/C3TQSF63C/p1732629758803629), I changed the `query`'s `body` and `description` to allow for an array of strings. We were already doing that for the `/articles` index, and as we intend on merging multiple fields within `description`, it makes sense for us to do it there as well. 
- Removed `quoteV2` from both `articles` transformers as we have none anymore and there was a TODO to remove it when that was the case.
- I had to change the snapshots tests helpers. This is because otherwise the tests got muddled with what is a document we fetch for an Addressable vs an document that's for the `events` and `articles` indexes. The latter expects much more information than the former. I hope the changes are clear. Basically the documents we fetch for `addressables` are now in a subfolder.

## How to test

`cd pipeline && yarn test`
`cd pipeline && yarn testTransformers --type=article`

Easier to test specific content types based on what their graph query fetches.

If all tests pass I think we're good.


pipeline/src/transformers/addressables/article.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const queryBody = data.body
+    ?.map(slice => {
+      if (['text', 'quote', 'standfirst'].includes(slice.slice_type)) {
+        return slice.primary.text.map(text => text.text);
+      } else {
+        return [];
+      }
+    })
+    .flat();
```
rcantin-w, line 41: I copied this from what we had in the base articles transformer, because we didn't need the content from embeds, or image galleries, etc. Just these three slices.

pipeline/src/transformers/addressables/book.ts
```diff
@@ -29,8 +29,7 @@ export const transformAddressableBook = (
     ?.map(s => {
       return s.primary.text.map(t => t.text);
     })
-    .flat()
-    .join(' ');
+    .flat();
```
rcantin-w, line 32: Now an array

pipeline/src/transformers/article.ts
```diff
@@ -112,10 +112,7 @@ export const transformArticle = (
 
   const queryBody = data.body
     ?.map(slice => {
-      if (
```
davidpmccormick, line 115: goooooood riddance

rcantin-w: (time of your life)

PR #171, opened 2024-11-27
Add transformer for addressable exhibition text

davidpmccormick:
## What does this change?

Adds a transformer for the exhibition texts, part of #156 

## How to test
`cd pipeline && yarn testTransformer --type exhibition-text`
image

We're closer to being able to use the content api for addressable content search

n/a


pipeline/src/graph-queries/addressables/exhibitionTexts.ts
```diff
@@ -2,6 +2,15 @@ const query = `
   exhibition-texts {
     title
     intro_text
+    related_exhibition {
```
davidpmccormick, line 5: had to add this to the query because the `intro_text` is often empty, and if it is, we show the `related_exhibition`'s promo caption so we have to fetch that too

pipeline/src/transformers/addressables/exhibitionText.ts
```diff
@@ -0,0 +1,43 @@
...
+  const primaryImage = isFilledLinkToDocumentWithData(data.related_exhibition)
+    ? data.related_exhibition.data.promo?.[0]?.primary
+    : undefined;
+
+  const promoCaption = primaryImage?.caption && asText(primaryImage.caption);
+
+  const description = introText || promoCaption || undefined;
+  const queryDescription = description ? [description] : undefined;
+
+  const title = asTitle(data.title);
```
rcantin-w, line 24: I think we said this would be "[name of exhibition] exhibition text)", so maybe that's an extra field to get from the related document?

rcantin-w: Although I think that's just for display (not query)

pipeline/src/transformers/addressables/exhibitionText.ts
```diff
@@ -0,0 +1,43 @@
...
+    id,
+    uid,
+    display: {
+      type: 'Exhibition text',
+      id,
+      uid,
+      title,
+      description,
+    },
+    query: {
```
rcantin-w, line 37: `body` is missing (Body: `guide_text_item` fields only)

davidpmccormick: thanks – totally missed this

rcantin-w (approved):
Looks good! I think the snapshots just need to be rerun for the test but once that's done and it passes, all good to go 👍

PR #172, opened 2024-11-28
Add addressable exhibition transformer

davidpmccormick:
## What does this change?

For #156 

Adding the addressable exhibition transformer

## How to test

`cd pipeline && yarn testTransformer --type exhibition`
image

Gets us closer to being able to use the content api for 'All' search

n/a

### DIFF
--- a/pipeline/src/transformers/addressables/exhibition.ts
+++ b/pipeline/src/transformers/addressables/exhibition.ts
@@ -1,0 +1,47 @@
+import {
+  asText,
+  asTitle,
+  isFilledLinkToDocumentWithData,
+  isNotUndefined,
+} from '@weco/content-pipeline/src/helpers/type-guards';
+import { ExhibitionPrismicDocument } from '@weco/content-pipeline/src/types/prismic';
+import { ElasticsearchAddressableExhibition } from '@weco/content-pipeline/src/types/transformed';
+
+export const transformAddressableExhibition = (
+  document: ExhibitionPrismicDocument
+): ElasticsearchAddressableExhibition => {
+  const { data, id, uid: documentUid } = document;
+  const format = isFilledLinkToDocumentWithData(data.format)
+    ? asText(data.format.data.title)
+    : 'Exhibition';
+
+  const dates = {
+    start: data.start,
+    end: data.end,
+  };
+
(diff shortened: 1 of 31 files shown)


pipeline/src/transformers/addressables/exhibition.ts
```diff
@@ -0,0 +1,47 @@
...
+    display: {
+      type: 'Exhibition',
+      id,
+      uid,
+      title,
+      description,
+      format,
+      dates,
+    },
+    query: {
```
rcantin-w, line 41: Format and contributors missing (We were thinking curators might be listed, or artists)

pipeline/src/transformers/addressables/exhibition.ts
```diff
@@ -0,0 +1,47 @@
...
+): ElasticsearchAddressableExhibition => {
+  const { data, id, uid: documentUid } = document;
+  const format = isFilledLinkToDocumentWithData(data.format)
+    ? asText(data.format.data.title)
+    : 'Exhibition';
+
+  const dates = {
+    start: data.start,
+    end: data.end,
+  };
```
rcantin-w, line 21: For the FE component, should we have the same data name for dates as Events (`times`) or would we treat them differently? Or maybe event should be `dates`, since there's no time 🤔

davidpmccormick: yeah, I wondered about this – I wasn't sure. I guess `dates` is slightly more generic and possibly a better fit even if we include things with times at some point it wouldn't be wrong?

pipeline/src/transformers/addressables/exhibition.ts
rcantin-w: It's why we do this! 💪

davidpmccormick: Looks like I'd added `format` as part of the `queryDescription` (since it isn't an allowed key on `AddressableQuery`). Will add contributors though.

rcantin-w: ah of course! thanks!

rcantin-w (approved):
`:finger-guns-doge:`

PR #173, opened 2024-12-02
Add addressable exhibition highlight tour transformer

davidpmccormick:
## What does this change?
Part of #156

## How to test
`cd pipeline &&  yarn testTransformer --type exhibition-highlight-tour`
 Screenshot 2024-12-02 at 17 23 34

We return an array of two objects from the one query – one for the audio guide and one for the BSL guide

We get closer to using the Content API for 'All' search

n/a


pipeline/src/transformers/addressables.ts
```diff
@@ -59,9 +64,10 @@ export const transformAddressable = (
       transformedDocument = transformAddressableExhibitionText(document);
       break;
 
-    // case 'exhibitions-highlight-tours':
-    //   transformedDocument = transformAddressableBook(document);
-    //   break;
+    case 'exhibition-highlight-tours':
+      transformedDocument =
+        transformAddressableExhibitionHighlightTour(document);
+      break;
```
rcantin-w, line 70: Did we figure out how to pass in an array, is it alright to do? We might have to change this slightly? https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/extractTransformLoad.ts

pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
```diff
@@ -0,0 +1,88 @@
...
+  const audioBody = getBody('audio');
+  const bslBody = getBody('bsl');
+
+  const uid = documentUid || undefined;
+
+  const shared = {
+    id,
+    uid,
+    display: {
+      type: 'Exhibition highlight tour' as const,
```
rcantin-w, line 64: Wonder why the `as const` is needed here 🤔

pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
```diff
@@ -0,0 +1,88 @@
...
+    : undefined;
+  const exhibitionTitle = exhibitionTitleField && asTitle(exhibitionTitleField);
+  const introText = data.intro_text && asText(data.intro_text);
+  const primaryImage = relatedExhibition
+    ? relatedExhibition.data.promo?.[0]?.primary
+    : undefined;
+
+  const promoCaption = primaryImage?.caption && asText(primaryImage.caption);
+
+  const description = introText ?? promoCaption;
```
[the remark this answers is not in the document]
rcantin-w: Ahhh I just clicked it's the Exhibition's itself! Sorry!

pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
```diff
@@ -0,0 +1,88 @@
...
+    uid,
+    display: {
+      type: 'Exhibition highlight tour' as const,
+      id,
+      uid,
+      description,
+    },
+    query: {
+      type: 'Exhibition highlight tour' as const,
+      description,
```
rcantin-w, line 71: Could query contain both intro text and promo caption?

rcantin-w (approved):
Looking great!

Just the question about what to do with the returned array, but maybe that's one for another ticket 🤷‍♀️

pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
davidpmccormick: I didn't dig into it, but maybe because of the spreads below

pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
davidpmccormick: I didn't dig into it but maybe because of the spreads below

pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
davidpmccormick: Good point – no, it wasn't supposed to. Will update.

pipeline/src/transformers/addressables.ts
davidpmccormick: I wasn't sure what was required, but I mostly ignored it once I got the tests passing. Maybe we can go through this together?

rcantin-w: Let's talk about it at the FE catchup!

pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
rcantin-w: [In Stories I put both the standfirst and the promo caption](https://github.com/wellcomecollection/content-api/pull/170/files#diff-d95d969e0baadd0614d2ff56d1784de3b351649b2331ec34a3f1eed9ae403981R59), so I thought maybe it could be a similar thing here, but maybe it's too repetitive..? 🤔

PR #176, opened 2024-12-03
Add addressable page transformer

davidpmccormick:
## What does this change?
For #156 adding the page transfomer

## How to test
`cd pipeline && yarn testTransformer --type page` 

We get closer to using our own APIs for 'All' search

n/a


rcantin-w (approved):
LGTM! I tested with a different page to check tags and description
Screenshot 2024-12-03 at 15 49 45

PR #179, opened 2024-12-04
Allow delisted items to be filtered from addressables query

davidpmccormick:
## What does this change?
We don't want to include event-schedule items in those that we're including in 'All' search. So we filter them out as early as possible (using `prismic.filter`)

## How to test
I'm not sure about this – don't think it's possible with our set up currently

We're not indexing event-schedule events that we don't want to display

The filter applies to the whole addressables query (because of the way we've grouped them). I think this is ok (can't think of anything that is `delist`ed that we'd want included in addressable search)


gestchild (approved):

rcantin-w (commented):
I thought about testing it with our test scripts (graphQuery and transformer) using `ZtXAxhEAACEAMXXK` (delisted event), thinking they should error/return nothing, but because they use a different `getById` it just works. Could/should we centralise getById so it uses the same config on the pipeline as in the tests?

PR #181, opened 2024-12-04
Return arrays from transformers for use with mergeMap

davidpmccormick:
## What does this change?
All transformers now always return an array, so that we can us Rx.js `mergeMap` (replaces `map`) consistently (ht @kenoir).

## How to test
`cd pipeline && yarn test`

I think this is the last piece required for the addressable transformers, so 'All' search will soon be possible.

A future reindex will include this.


pipeline/src/transformers/addressables.ts
```diff
@@ -28,7 +28,31 @@ import { transformAddressableExhibitionText } from './addressables/exhibitionTex
 import { transformAddressablePage } from './addressables/page';
 import { transformAddressableVisualStory } from './addressables/visualStory';
 
-export const transformAddressable = (
+export function transformAddressable(
```
davidpmccormick, line 31: I rewrote this using [overload signatures](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) so that the return type is known based on the input parameter type

gestchild:
I've merged the season and project transformers into main, so will need to amend them to return arrays. Do you mind doing that in this PR?

rcantin-w (approved):
Just had a look through and tested and it looks good - looking forward to testing with a new pipeline 💪

PR #183, opened 2024-12-06
Extent next opening dates list

agnesgaroux:
## What does this change?

With the extended freeze discussed with LE&E, the list of days needs to be longer so that we're still left with at least 12 options once we remove all the closed days

## How to test

Deepstore requests will be unavailable from 14 Dec to 1 Jan (19 days) 
add 14 days lead time (10 working days + 4 weekends closed day) 
add 14 days to choose from (12 library opens days + 2 Sundays)
=> 47 days needed in the initial list
Plus a few extra just to be safe

No complaints from LE&E 

Requesting can be turned off as a last resort


api/src/controllers/utils.ts
```diff
@@ -15,7 +15,7 @@ export function getNextOpeningDates(
   // 35 is somewhat arbitrary,
   // we just need to have at least 12 once all the closed days and/or delivery time have been removed
   const dateNow = new Date();
-  const dateList = [...Array(35).keys()].map(day => addDays(dateNow, day));
```
kenoir, line 18: Do we need to update the comment too?

kenoir (approved):

api/src/controllers/utils.ts
agnesgaroux: Updated, along with the one further below about formatting the exceptional closed days

Issue #167, opened 2024-11-26
Add `/addressables` endpoint to API and add tests


rcantin-w:
Ensure the new endpoint is added to our [API app file](https://github.com/wellcomecollection/content-api/blob/main/api/src/app.ts) and add tests.

Issue #166, opened 2024-11-26
Create API query for Addressables


rcantin-w:
Create the "query" file for Addressables ([see articles'](https://github.com/wellcomecollection/content-api/blob/main/api/src/queries/articles.ts)). This will determine what we query and how we score them. This could be kept quite simple as we don't need filters or aggregations, and we kept it very simple for the `events` endpoints. All exercises in improvement relevancy can be taken on at a later point.

PR #186, opened 2024-12-09
Add addressables API query

davidpmccormick:
For #166 

## What does this change?

Adds the API query for addressables. It doesn't include any filtering or aggregation, in line with what was [decided in the RFC](https://github.com/wellcomecollection/docs/tree/main/rfcs/062-content-api-all-search#background-information).

## How to test
This query isn't connected to a controller (coming in #167) so doesn't require testing yet because it isn't doing anything.

n/a

n/a

### DIFF
--- a/api/src/queries/addressables.ts
+++ b/api/src/queries/addressables.ts
@@ -1,0 +1,21 @@
+import { QueryDslQueryContainer } from '@elastic/elasticsearch/lib/api/types';
+
+export const addressablesQuery = (
+  queryString: string
+): QueryDslQueryContainer => ({
+  multi_match: {
+    query: queryString,
+    fields: [
+      'id',
+      'uid',
+      'query.title.*^100',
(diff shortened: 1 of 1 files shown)


api/src/queries/addressables.ts
```diff
@@ -0,0 +1,21 @@
+import { QueryDslQueryContainer } from '@elastic/elasticsearch/lib/api/types';
+
+export const addressablesQuery = (
+  queryString: string
+): QueryDslQueryContainer => ({
+  multi_match: {
+    query: queryString,
+    fields: [
+      'id',
+      'uid',
```
davidpmccormick, line 10: I added `uid` which doesn't currently exist in the other queries. I wasn't sure whether we would want to add `type` here too (since it might help determine what we want to score higher amongst the various addressable types)

rcantin-w (approved):

PR #187, opened 2024-12-09
Add addressables controller

davidpmccormick:
For #168 and #167

## What does this change?
Uses the API query in #186 to create controller for `/all` endpoint

## How to test
`cd api && yarn test`

We can return addressable content from the Content API

This was largely a process of copying and removing parts from another controller. Something might have got overlooked in the process


api/src/app.ts
```diff
@@ -20,6 +21,7 @@ const createApp = (clients: Clients, config: Config) => {
 
   app.use(morgan('short', { stream: logStream('http') }));
 
+  app.get('/all', addressablesController(clients, config));
```
rcantin-w, line 24: Doooo we want to keep this as `/addressables` or not? 
I don't have a strong opinion on this.

davidpmccormick: It was `/all` in the RFC? I think that would be my preference because it describes better what the API is returning rather than what it has indexed (which doesn't necessarily make much sense to a consumer of the API without knowing its inner workings). I think I've just largely reverse engineered that opinion though

rcantin-w: (and `all` in the pipeline means reindex all indices, but you're right that we can do the distinction at that point)

api/src/controllers/addressables.ts
```diff
@@ -0,0 +1,73 @@
...
+import { addressablesQuery } from '@weco/content-api/src/queries/addressables';
+import { Clients, Displayable } from '@weco/content-api/src/types';
+import { ResultList } from '@weco/content-api/src/types/responses';
+
+import { HttpError } from './error';
+import { paginationElasticBody, PaginationQueryParameters } from './pagination';
+
+type QueryParams = {
+  query?: string;
+  'contributors.contributor'?: string;
```
rcantin-w, line 17: I think contributors could be removed?

rcantin-w (approved):
tests are passing - we'll be able to manually test once we've indexed but it looks good!

PR #191, opened 2025-01-06
Move caption/description logic into a helper

davidpmccormick:
## What does this change?

We had a lot of repeated code to get the caption from a promo image. This adds a helper to get the caption.

## How to test

`cd pipeline && yarn test` – snapshots still pass

We have fewer places for bugs to be introduced and a more straightforward path for updating the logic in future if necessary.

This is the first time I've looked at a code editor since 2024.

### DIFF
--- a/pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
+++ b/pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
@@ -22,9 +22,8 @@ export const transformAddressableExhibitionHighlightTour = (
     : undefined;
   const exhibitionTitle = exhibitionTitleField && asTitle(exhibitionTitleField);
   const introText = data.intro_text && asText(data.intro_text);
-  const promoCaption = primaryImageCaption(relatedExhibition?.data.promo);
-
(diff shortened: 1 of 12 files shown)


rcantin-w (approved):

pipeline/src/transformers/addressables/exhibitionHighlightTour.ts
```diff
@@ -21,11 +22,7 @@ export const transformAddressableExhibitionHighlightTour = (
     : undefined;
   const exhibitionTitle = exhibitionTitleField && asTitle(exhibitionTitleField);
   const introText = data.intro_text && asText(data.intro_text);
-  const primaryImage = relatedExhibition
-    ? relatedExhibition.data.promo?.[0]?.primary
-    : undefined;
-
-  const promoCaption = primaryImage?.caption && asText(primaryImage.caption);
+  const promoCaption = primaryImageCaption(relatedExhibition?.data.promo);
```
rcantin-w, line 25: Would it be sensible to align them all to be `promoCaption` (also named `description` and `caption` in other transformers).

Issue #195, opened 2024-12-16
Add e.g. `displayType` to Exhibition Highlight Guide display block


davidpmccormick:
We don't currently have a way to distinguish between `bsl` and `audio` exhibition highlight tours on the front end from the Content API. We'd need this to be able to show users an appropriate label, as well as to be able to create the appropriate link to either the bsl or audio guide page.

We concluded that adding another property to the `display` object would likely be the most straightforward solution.

rcantin-w:
@pollecuttn this was deployed yesterday: https://api.wellcomecollection.org/content/v0/all?query=highlight-tour, we changed the name from `displayType` to `highlightTourType`.

<img width="681" alt="Image" src="https://github.com/user-attachments/assets/41511afc-4edc-48aa-860e-6d01d2d0bdb7" />

PR #196, opened 2025-01-08
Add displayType to exhibition highlight tours

davidpmccormick:
## What does this change?

For #195 – including an extra property in the `display` block for exhibition-highlight-tours so that we can indicate whether it is about BSL or audio, and be able to link to the right place from the front end.

## How to test

`cd pipeline && yarn test` – tests pass (snapshot has been updated)

Users understand what a search result is about and can be linked to the appropriate content from the search result

n/a


rcantin-w:
Screenshot 2025-01-08 at 16 49 34


As we already have `type`, is `displayType` the name we want to use/explains what this is for? Maybe `highlightTourType`? Or something better 😅

rcantin-w:
Tests all run well locally 👍

rcantin-w (approved):

gestchild:
> As we already have `type`, is `displayType` the name we want to use/explains what this is for? Maybe `highlightTourType`? Or something better 😅

format?

PR #197, opened 2025-01-14
Add audiences and format to transformer for query field.

rcantin-w:
## What does this change?

- Adds audiences and format to events query object in transformer, as an array of string (audiences labels) and single string (format label).
- Added them to the mapping
- Added them to the query, kind of copied from what was done in articles. I suggest going ahead with that and amending should we think a change is needed.

## How to test

Run api locally and search for http://localhost:3000/events?query=The%20Hub, which is a series' contributors. These are currently indexed in the query object but not part of what is considered relevant. Locally, it now is. So you can see it should return more results than live: https://api-stage.wellcomecollection.org/content/v0/events?query=The%20Hub

Results are more useful and more fields return results.

Updating the mapping will be required before running a reindex. I've never done it but have tested locally and _adding fields_ to the mapping should be straightforward.


davidpmccormick (approved):
Makes sense. Looks good.

rcantin-w:
Converting to draft as dependant on new pipeline being created

PR #199, opened 2025-01-20
Pipeline 2025-01-20

rcantin-w:
## What does this change?

- Adds new module for new pipeline and a brand new stack repo for it
- Adds a README in `infrastructure` on how to add/delete pipelines
- Adds an endpoint (`/_elasticConfig`) that returns the pipeline date that's in use
  

## How to test
- Terraform plan should only show changes on creating a new pipeline
- Accessing http://localhost:3000/_elasticConfig returns what pipeline date is currently the one being used

Able to move forward with future work

N/A


infrastructure/pipeline_2025-01-20_stack/providers.tf
```diff
@@ -0,0 +1,18 @@
...
+    ec = {
+      source = "elastic/ec"
+    }
+    elasticstack = {
+      source = "elastic/elasticstack"
+    }
+  }
+}
+
+provider "elasticstack" {
```
kenoir, line 12: So long as this provider is within the module there will be issues deleting instances of it without a targeted destroy. Unfortunately because provisions elasticsearch users seems to require its own provider we may be stuck with this problem. 

There's no need to create a copy of the pipeline stack module without any changes, in which case just create another instance of the existing module.

rcantin-w:
@kenoir after our conversation I've removed the duplicates and am using the default stack in the new pipeline. I've also added a README, which I'd love your review on, which explains how to create, work on, and destroy pipelines.

Issue #201, opened 2025-01-23
Delete 2024-12 pipeline


rcantin-w:
Leaving it over the weekend to see how the new pipeline is performing, then we should delete it ASAP to save on costs - thinking Monday.

rcantin-w:
Done

PR #202, opened 2025-01-27
Update README on how to delete a pipeline and remove old pipeline module

rcantin-w:
## What does this change?

#201 

Removes the old pipeline as the API [uses the new one](https://api.wellcomecollection.org/content/v0/_elasticConfig) since Friday and no problems have come up.

I've already destroyed the old pipeline with a targeted destroy so this just ensure `main` is up-to-date, as well as updates the README instructions based on what I ran into when doing this work!

## How to test
Does the README make sense? 
Does running `terraform plan` flag any changes?

We now have everything we need to create/destroy pipelines on our own next time.

N/A already applied


gestchild (approved):
Thanks for the README - very helpful

PR #205, opened 2025-01-31
Update ports to use local nginx for dev

kenoir:
## What does this change?

Part of https://github.com/wellcomecollection/wellcomecollection.org/pull/11551

This change updates the port that content API uses when run locally in order that it can run at the same time locally to enable a more complete local development experience when working with https://github.com/wellcomecollection/wellcomecollection.org

## How to test

- In the `./api` folder run:
  - `yarn && yarn dev`

The content API should be available on `localhost:3002`.

This change can be more thoroughly tested by reviewing https://github.com/wellcomecollection/wellcomecollection.org/pull/11551 and following the relevant steps described running this branch.

Easier development!

This should have minimal impact as there are no prod changes, and the correct port is flagged to devs when the app starts.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "start": "NODE_ENV=production tsx ./server.ts",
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
Works well by itself, having a look at the wc.org PR now

Issue #121, opened 2024-03-19
Align articles/events filters on API faceting principles


agnesgaroux:
# NEW TICKET DESCRIPTION
After discussing this further, we decided the former ticket description wasn't exactly what was needed, but keeping it for legacy conversations.

See comments in ticket for more information as to what was decided upon, [starting here](https://github.com/wellcomecollection/content-api/issues/121#issuecomment-2633614082).

# FORMER TICKET DESCRIPTION
https://github.com/wellcomecollection/docs/tree/main/rfcs/037-api-faceting-principles
See Principle 1. 
> Filters are named by the JSON paths of the identified object that they filter or, if applied to an attribute other than the identifier, the path of that attribute

This is not something that has been abided by in content-api: 
NOTE: this example is articles but same thing with events
Article doc `XkvmsREAACUAUWwN` has a filter object like
```
        "filter": {
            "contributorIds": [ // should be "contributors.contributor"
              "XjQP8BQAACQANdYu",
              "XjFs8xQAACMAKkcQ"
            ],
            "formatId": "W7TfJRAAAJ1D0eLK",  // should be "format"
            "publicationDate": "2020-02-27T11:31:56.000Z"
```
and display 
```
 "display": {
      "type": "Article",
      "id": "XkvmsREAACUAUWwN",
      "title": "The ‘undesirable epileptic’",
      "caption": "Abused in her marriage for being 'a sick woman', Aparna Nair looked to history to make sense of the response to her epilepsy. 
She discovered how centuries of fear and discrimination were often endorsed by science and legislation.",
      "format": {
        "type": "ArticleFormat",
        "id": "W7TfJRAAAJ1D0eLK",
        "label": "Article"
      },
      "publicationDate": "2020-02-27T11:31:56+0000",
      "contributors": [
        {
          "type": "Contributor",
          "contributor": {
            "type": "Person",
            "id": "XjQP8BQAACQANdYu",
            "label": "Aparna Nair"
          },
          "role": {
            "type": "EditorialContributorRole",
            "id": "WcUWeCgAAFws-nGh",
            "label": "Author"
          }
        },
        {
          "type": "Contributor",
          "contributor": {
            "type": "Person",
            "id": "XjFs8xQAACMAKkcQ",
            "label": "Tracy Satchwill"
          },
          "role": {
            "type": "EditorialContributorRole",
            "id": "XVQa_xMAACMAqbwo",
            "label": "Illustrator"
          }
        }
      ],
    }
```

[...]

PR #206, opened 2025-02-06
Align filters names for events and articles

rcantin-w:
> [!WARNING]
> DO NOT MERGE THIS UNTIL WE ARE READY TO GO AHEAD WITH THE NEW PIPELINE.
> This changes the mapping and the transformer, which won't work with the current pipeline.

Added Paul as a reviewer as he's the consultant on the work; but @wellcomecollection/digital-experience approval will be necessary too ✨ 

## What does this change?

#121 
Removes the notion of "Ids" in filter names
See ticket for more information, [starting at this comment](https://github.com/wellcomecollection/content-api/issues/121#issuecomment-2633614082).

## How to test

If you run it locally, passing in query params, does the API still return the same thing as production?

Aligned with our principles and better to be built upon.

It breaks search! But we'll be able to test this locally and in staging first.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,5 +1,5 @@
 env:
-  LIVE_PIPELINE: '2025-01-20'
+  LIVE_PIPELINE: '2025-02-06'
 steps:
   - label: 'autoformat'
     command: '.buildkite/scripts/autoformat.sh'
--- a/api/config.ts
+++ b/api/config.ts
@@ -12,7 +12,7 @@ const environment = environmentSchema.parse(process.env);
 // This configuration is exposed via the public healthcheck endpoint,
 // so be careful not to expose any secrets here.
 const config = {
-  pipelineDate: '2025-01-20',
+  pipelineDate: '2025-02-06',
   addressablesIndex: 'addressables',
   articlesIndex: 'articles',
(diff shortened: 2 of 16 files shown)


davidpmccormick (approved):
I have a vague (probably wrong) memory of us talking about everything being singular rather than plural (e.g. `contributor` instead of `contributors`), but here `format` is the only singular one – is that right? I possibly misunderstood something

rcantin-w:
@davidpmccormick I posted a lot in [the ticket](https://github.com/wellcomecollection/content-api/issues/121) about this and would love your opinion as it felt very "🤷‍♀️ this should do", when you have time! Format is singular as a document can only have one, but we did have conversations about this with Paul. His comments are in the ticket thread as well! Let me know what you think 🙏

davidpmccormick:
things that can be arrays being plural works for me (definitely what I would have done without any external input), so as long as Paul's happy then all good.

rcantin-w:
Approved but converting to draft to avoid caffeine-lacking-brains merging it since it would affect the existing pipeline. Can't do it until the work [here](https://github.com/wellcomecollection/content-api/pull/209) and [here](https://github.com/wellcomecollection/content-api/pull/210) are also good to go, at which point we'll move to using the new pipeline.

PR #210, opened 2025-02-10
Interpretations query

rcantin-w:
## What does this change?

#11593

Adds `interpretations` to mapping of the events index and to its query score.
I put it at the same level as `audiences`.

Before
Screenshot 2025-02-10 at 10 27 59

After
Screenshot 2025-02-10 at 10 28 20

## How to test

- Run the API locally pointing at the new pipeline (`2025-02-06`) and query for interpretations labels, see that relevant results are returned. 
- You could run .org locally pointing at the API pointing at the new pipeline to see the results returned on the cards as well (as per the images above)

More results when searching for accessibility things!

More results = more noise = less relevant results? We'll be working on that.


gestchild (approved):
looks good and I can see the increased results when pointing at the new pipeline

Issue #193, opened 2025-01-07
Add date filtering to Events; build the logic


rcantin-w:
## Definition of ready
- [x] Questions from #130 have been answered

## Context
Read [this RFC](https://github.com/wellcomecollection/docs/tree/main/rfcs/061-content-api-next-steps#challenges-encountered-and-decisions-made) first to get acquainted with the challenges met on this piece of work 
Read #130 for final decisions.

Designs
<img width="746" alt="Image" src="https://github.com/user-attachments/assets/c306da99-e991-463c-a81f-798e45e5dec0" />

## Definition of done

- [ ] Can query by the filters above on the API
- [ ] Only one filter can be used at a time (error message otherwise)

rcantin-w:
@LaurenFBaily the new pipeline is live - I'll wait until next week to delete the old one in case something happens and we need to revert.

So if you go to http://api.wellcomecollection.org/content/v0/events?timespan=this-week and change "this-week" to [this selection](https://github.com/wellcomecollection/content-api/pull/215/files#diff-5782f3a6fece4e5ecf33e06b90c78f07a7f3802669da9c402893a5c4be927408R75-R96), the results should make enough sense - next step will be to test it with mock events - should we book something in for next week?

If you'd like a more visual way to see what comes up, it will work on the FE too, so https://wellcomecollection.org/search/events?timespan=this-week

PR #215, opened 2025-02-13
Add timespan to events filter object

rcantin-w:
## What does this change?

#193 

> [!NOTE]
> This PR is not the final work, but having this pipeline become the live one will help us test what we need to do next. Merging this **should not change anything on the UI** (the filter should not appear), if it does, do flag! But being able to enter the query manually will be super useful so it'd be nice to get it through.

- Points to new pipeline (2025-02-17) as I needed to change the mapping to include `timespan` in the eventDocuments' filter object.
- Added type guards, validators and many helpers for monthly filtering and the likes.
- The timespan had to be passed somewhere different in the actual Elasticsearch query. This is partly why adding this filter was a bit more complex, I couldn't just copy paste the way we did it elsewhere. Please do review the code and let me know if anything is _unclear_ and could benefit from comments left in the code!
- Added a `getTimespanRange` helper function which returns what we need for each query. _THIS IS NOT THE FINAL VERSION OF THEM_. We need further testing and to create mock events to test with. See https://github.com/wellcomecollection/content-api/issues/221. If you _do_ spot things are WAY OFF, do flag it's appreciated! It's a complex thing 🙏 
- Added tests for `getTimespanRange` that are hopefully readable and put things into context, but maybe they could be more useful? Let me know what you think. I'd definitely like to test the query itself further and check for specific results from the past, for example. It's a WIP!

## How to test

- Run locally and play around the API responses. You could also run the FE and test it there if you'd rather visual results! 
- Do tests pass? Should more tests be added?
- Does it change the UI at all on the FE? **It shouldn't.**

Way easier to test for next steps for Lauren and Dana, and me!

Breaks something 👀


gestchild (approved):
I've run everything locally and all looks good. The tests all pass.
I think everything makes sense and the tests are clear. ✨

PR #227, opened 2025-02-26
Add pipeline `2025-02-26` , creation script and modify README

rcantin-w:
## What does this change?

The README instructions weren't 100% working (couldn't set required variables and then run scripts, they had to be declared every time). So I also created a bash script to create pipelines and removed the instructions from the README.

Basically I don't really know what I'm doing with this, it's my first time! I tried to put in some validation, but I'd welcome any improvement suggestions/corrections/telling off about it! Thank you!

## How to test

Create a new pipeline I guess? I've run it locally and it does the job, happy to help with the reviewing/testing process.

Easier/faster to create new pipelines.

Again it's my first time with this, so potentially there are risks, yes.

### DIFF
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -3,37 +3,12 @@
 ## Creating pipeline
 
 1. Add new module to `pipeline.tf`, named with the date.
-2. From `./infrastructure`, plan and apply Terraform (`terraform init` might be required first)
-
-At this stage you could chose to do it manually (see instructions below), or run `./create_pipeline.sh`. If you chose to do so, make sure you run it with its required arguments; `./create_pipeline.sh PIPELINE_DATE=[NEW PIPELINE DATE] AWS_PROFILE=catalogue-developer BUILDKITE_COMMIT=dev`.
-
+2. From `./infrastructure`, plan and apply Terraform (`./run_terraform.sh plan`), although `terraform init` will probably be required first.
 3. Now your pipeline should have been created, you may confirm so by going into AWS Lambdas as well as ElasticCloud and seeing that it's there.
+4. Run `./create_pipeline.sh [NEW PIPELINE DATE]`.
+5. The Elasticsearch indexes will have been created and configured, but you'll need to run a reindex manually. The scheduler will be set but only considers the last 15 minutes of documents being published, so the inital run has to be manual (although it could be made to be automated should we desire it). Before you reindex, **_remember to change `pipelineDate` values locally to match your new pipeline's._**, otherwise you'll be doing those in the live pipeline.
 
-You need to actually fill the Lambdas with your code. For all the steps below, **MAKE SURE YOU ARE SPECIFYING THE NEW PIPELINE'S NAME**, otherwise the default will apply to the live one.
-From the project root, run command lines from the `publish: pipeline & unpublisher ($LIVE_PIPELINE)` task in `.buildkite/pipeline.yml` locally, skipping `yarn install`:
-
-1. The first two lines are creating zips of the required packages, you can run them as is:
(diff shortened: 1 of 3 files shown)


infrastructure/create_pipeline.sh
```diff
@@ -0,0 +1,30 @@
...
+  echo "ERROR: PIPELINE_DATE, BUILDKITE_COMMIT and AWS_PROFILE are mandatory arguments.";
+  exit 1;
+fi
+
+echo Create zips for packages
+yarn workspace @weco/content-pipeline run package && yarn workspace @weco/content-unpublisher run package
+
+echo Upload zips to S3
+bash ../.buildkite/scripts/upload_lambda_package.sh content-pipeline-$PIPELINE_DATE ../pipeline/package.zip $AWS_PROFILE=$AWS_PROFILE $BUILDKITE_COMMIT=$BUILDKITE_COMMIT
+bash ../.buildkite/scripts/upload_lambda_package.sh content-unpublisher-$PIPELINE_DATE ../unpublisher/package.zip $AWS_PROFILE=$AWS_PROFILE $BUILDKITE_COMMIT=$BUILDKITE_COMMIT
```
kenoir, line 25: The shebang (`#!/usr/bin/env bash`) at the top of the script tells the shell how to execute this script if the script itself is marked as executable. So you don't need to reference `bash` here if those scripts have a shebang.

I'm not sure setting the $AWS_PROFILE=$AWS_PROFILE is working as expected. I think you may want to export this environment variable in the script before executing the other scripts so that it is available in the process of the child scripts. This is _already done_ when you set `AWS_PROFILE=catalogue_developer` above.

`$BUILDKITE_COMMIT=$BUILDKITE_COMMIT` is not used by the consuming script AFAICT and can be omitted.

It's also a good idea to use absolute paths by some known reference if possible to avoid things not being where you expect if the script is run from an unexpected location, so we use `ROOT` as defined above.

```suggestion
$ROOT/.buildkite/scripts/upload_lambda_package.sh \
  content-pipeline-$PIPELINE_DATE $ROOT/pipeline/package.zip
$ROOT/.buildkite/scripts/upload_lambda_package.sh \
  content-unpublisher-$PIPELINE_DATE $ROOT/unpublisher/package.zip 
```

infrastructure/README.md
```diff
@@ -4,28 +4,29 @@
 
 1. Add new module to `pipeline.tf`, named with the date.
 2. From `./infrastructure`, plan and apply Terraform (`terraform init` might be required first)
+
+At this stage you could chose to do it manually (see instructions below), or run `./create_pipeline.sh`. If you chose to do so, make sure you run it with its required arguments; `./create_pipeline.sh PIPELINE_DATE=[NEW PIPELINE DATE] AWS_PROFILE=catalogue-developer BUILDKITE_COMMIT=dev`.
```
kenoir, line 8: I would avoid having both a set of instructions and a script to avoid one diverging without the other. If there's a script that works, we can point at that and add comments inline that document what it's doing.

infrastructure/create_pipeline.sh
rcantin-w: BUILDKITE COMMIT is used here, yes? https://github.com/wellcomecollection/content-api/blob/main/.buildkite/scripts/upload_lambda_package.sh#L10

rcantin-w: Shall I just make it `dev`, is there a reason we'd want a different one?

rcantin-w: Spoke in person, needed to export the value and has been set to dev explicitely

infrastructure/README.md
rcantin-w: Thank you for all your comments, I've made all the changes now!

kenoir (approved):

PR #229, opened 2025-02-27
Further timespan work

rcantin-w:

### DIFF
--- a/pipeline/src/transformers/eventDocument.ts
+++ b/pipeline/src/transformers/eventDocument.ts
@@ -270,7 +270,7 @@ export const transformEventDocument = (
   const isAvailableOnline = !!availableOnline;
 
   // If an event has scheduled times, we don't want to have the parent's time range in the filterable times.
-  const filterTimes = scheduledTimes || times;
+  const filterTimes = scheduledTimes.length > 0 ? scheduledTimes : times;
 
   return [
     {
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,5 +1,5 @@
 env:
-  LIVE_PIPELINE: '2025-02-17'
+  LIVE_PIPELINE: '2025-02-26'
 steps:
   - label: 'autoformat'
     command: '.buildkite/scripts/autoformat.sh'
--- a/api/config.ts
+++ b/api/config.ts
@@ -12,7 +12,7 @@ const environment = environmentSchema.parse(process.env);
 // This configuration is exposed via the public healthcheck endpoint,
 // so be careful not to expose any secrets here.
 const config = {
-  pipelineDate: '2025-02-17',
+  pipelineDate: '2025-02-26',
   addressablesIndex: 'addressables',
   articlesIndex: 'articles',
   eventsIndex: 'events',
--- a/api/scripts/holiday_closure_test.ts
+++ b/api/scripts/holiday_closure_test.ts
@@ -69,7 +69,7 @@ const applyItemsApiDeepstoreLogic = (
 const run = async () => {
   const elasticClient = await getElasticClient({
     serviceName: 'api',
-    pipelineDate: '2025-02-17',
+    pipelineDate: '2025-02-26',
     hostEndpointAccess: 'public',
   });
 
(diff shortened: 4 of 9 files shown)


pipeline/src/transformers/eventDocument.ts
```diff
@@ -270,7 +270,7 @@ export const transformEventDocument = (
   const isAvailableOnline = !!availableOnline;
 
   // If an event has scheduled times, we don't want to have the parent's time range in the filterable times.
-  const filterTimes = scheduledTimes || times;
+  const filterTimes = scheduledTimes.length > 0 ? scheduledTimes : times;
```
rcantin-w, line 273: scheduledTimes always returned _something_ (`[]`) so the logic was flawed

pipeline/src/transformers/eventDocument.ts
```diff
@@ -139,7 +139,7 @@ const transformTimes = ({
     .map((time): EventDocumentTime | undefined => {
       return {
         startDateTime: asDate(time.startDateTime) || undefined,
-        endDateTime: asDate(time.endDateTime) || undefined,
+        endDateTime: asDate(time.endDateTime) || new Date(2100, 1, 1),
```
rcantin-w, line 142: Undefined will not work for our range check, so if an event (most likely a permanent exhibition) does not have an end date, we assign one way in the future.

api/src/controllers/utils.ts
```diff
@@ -134,43 +131,29 @@ export const getTimespanRange = (
...
               lte: now.startOf('week').plus({ days: 6 }).endOf('day'), // Sunday
-              relation: 'contains',
             },
           },
         },
         {
           range: {
             'filter.times.endDateTime': {
-              gt: 'now',
+              gt: isNowWeekend ? 'now' : friday5PM, // Friday 5pm or NOW
```
davidpmccormick, line 145: nit: I'd say this comment isn't necessary because the variables are descriptive enough

api/src/controllers/utils.ts
```diff
@@ -207,8 +189,8 @@ export const getTimespanRange = (
       return [
         {
           range: {
-            'filter.times.startDateTime': {
-              gte: 'now',
+            'filter.times.endDateTime': {
```
davidpmccormick, line 192: this change seems like it's deliberate, but just checking it's definitely what we want?

davidpmccormick (approved):
Looks good

api/src/controllers/utils.ts
rcantin-w: Yes! So it says as long as the **end** date is greater than now, consider it the future, because it's not ended yet.

The previous query said "if the start date if greater or equal to now it's future", but an exhibition that started the previous month would then not have been returned.

gestchild:
Capturing the issues we discussed:

- ?timespan=all returns a 500 error, but we don't need it so should remove it as a [possible option](https://github.com/wellcomecollection/content-api/blob/9affa1fc8c93bfbcbcbcbdce8c49159ebfb074c6/api/src/controllers/events.ts#L90)
- this-week and may not returning all the possible results. I think we said the missing events had times missing in elastic

Not for this PR:
- ordering of events needs consideration, taking into account whether a query is present or not

gestchild:
The other thing that occurred to me while looking at this was whether we should show the dates on the cards rather than the past indicator, when you've filtered by past

rcantin-w:
Thank you @gestchild 🙌 

- `?timespan=all` returns a 500 error, but we don't need it so should remove it as a [possible option](https://github.com/wellcomecollection/content-api/blob/9affa1fc8c93bfbcbcbcbdce8c49159ebfb074c6/api/src/controllers/events.ts#L90) **addressed!**
- this-week and may not returning all the possible results. I think we said the missing events had times missing in elastic. **The fix [is here](https://github.com/wellcomecollection/content-api/pull/229/files#diff-b7c3709a3cb305505cb58589b163d3d8bae187d734c817f43383b3a8936f5c1fR273) but the reindexes got it. We tested again and it looks good!.**
- ordering of events needs consideration, taking into account whether a query is present or not **[Ticket created](https://github.com/wellcomecollection/content-api/issues/230).**
- The other thing that occurred to me while looking at this was whether we should show the dates on the cards rather than the past indicator, when you've filtered by past. **I feel like we discussed this @LaurenFBaily and @dana-saur but can't fully recall the outcome?**

gestchild (approved):

Issue #231, opened 2025-03-10
Delete `pipeline_2025-02-17` by EoD 13/03


rcantin-w:
Live pipeline switched to 02-26 as part of https://github.com/wellcomecollection/content-api/pull/229

After a couple days we can delete the old one.

PR #234, opened 2025-03-12
Remove pipeline_2025-02-17

davidpmccormick:
For #231 

## What does this change?
Deletes the old pipeline

## How to test
n/a

Old modules aren't hanging around

n/a
image


rcantin-w (approved):
🙌

PR #242, opened 2025-04-22
Add validation on Prismic filter values as well as dates

rcantin-w:
## What does this change?

We were getting [a lot of errors in the alerts channel](https://wellcome.slack.com/archives/CQ720BG02/p1744963470565019) because bad values were passed in the Content API's filters (e.g. `1'"` as an audience...), which returns a 500 and logs in the channel.

I added a validator for Prismic Ids (which also account for the few uuids we pass in the Content API), which will at least check that only alphanumerical values are passed in and return a more useful error message as "Bad Request". 

I also added a paramsValidator to `articles`, including a date checker (for `YYYY-MM-DD`, which according to tests (query.test.ts) is what we want from a date). We don't currently use Dates as a filter in articles, but we'll be covered when we do.

Results in;
Screenshot 2025-04-22 at 16 29 55

The FE still returns it as a "Content API Error", which we could address should we want to - do we? (I feel like; not necessarily, why would normal users come across these?)
Screenshot 2025-04-22 at 16 47 43

## How to test

Run locally bad requests (take inspo from the alerts channel link above!)
Run good requests, they should be unaffected.

Less "expected fails" in the alerts channel, making it more useful.

N/A?

### DIFF
--- a/api/src/controllers/articles.ts
+++ b/api/src/controllers/articles.ts
@@ -22,7 +22,12 @@ import { ResultList } from '@weco/content-api/src/types/responses';
 
 import { HttpError } from './error';
 import { paginationElasticBody, PaginationQueryParameters } from './pagination';
-import { queryValidator, validateDate } from './validation';
+import {
+  dateValidator,
+  prismicIdValidator,
+  queryValidator,
+  validateDate,
+} from './validation';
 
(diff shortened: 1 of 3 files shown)


gestchild (approved):
tested locally and works as expected

Issue #238, opened 2025-04-02
Allow filtering out


LaurenFBaily:
We'll have to add capacity to filter OUT in order to be efficient in for `/events` (removing the format "Exhibition"). We currently filter BY only.

Discuss with Platform if they have the capacity to do so in an effort to stay aligned. 

It could be ideal to do this for all listing endpoints, but not necessary.

### Done when
- [ ] Filtering out works 🤷‍♀ 
- [ ] Tests have been added.

PR #245, opened 2025-04-29
Filter out

gestchild:
## What does this change?

For [#238](https://github.com/orgs/wellcomecollection/projects/18/views/1?pane=issue&itemId=104671543&issue=wellcomecollection%7Ccontent-api%7C238)

~Allows us to filter out events based on their format, e.g. excludeFormats=Study%20day, Late, Workshop~
Allows us to filter our events with a format of 'exhibition' using the [id we add in the events transformer](https://github.com/wellcomecollection/content-api/commit/6aee340a5f735b209ed616151c7fead47f19747f#diff-90c5c725563bbc14ecb7d2e9ae4ee9709f41db4211882f694f2445b8cc65107eR17)

We'll use this to allow us to filter out events with a format of 'Exhibition' on the /events listing page

N.B. Would appreciate feedback on the naming of the exclude param in the url, or if people can think of a better way of doing it

## How to test
- visit http://localhost:3002/events
- append `?filterOutExhibitions=true`
- Nothing should change as we have no events with the exhibition format id.
- You can change the id in api/src/controllers/events.ts to 'WmYRpCQAACUAn-Ap' (the one for Gallery tour) and see the number of results change
- play with multiple formats combined with queries etc. check the results are what is expected.

We are able to retrieve all the events without including those with an exhibition format

- There may be a more efficient way to make the query to Elasticsearch

### DIFF
--- a/api/src/controllers/events.ts
+++ b/api/src/controllers/events.ts
@@ -170,6 +170,12 @@ const eventsController = (clients: Clients, config: Config): EventsHandler => {
       pick(eventsAggregations, requestedAggs)
     );
 
+    const excludes = pickFiltersFromQuery(
+      ['excludeFormats'],
+      validParams,
+      eventsExclusionsFilter
+    );
+
     const postFilters = pickFiltersFromQuery(
       [
         'format',
--- a/api/src/controllers/events.ts
+++ b/api/src/controllers/events.ts
@@ -89,7 +89,7 @@ const timespanValidator = queryValidator({
 });
 
 const paramsValidator = (params: QueryParams): QueryParams => {
-  const { isAvailableOnline, ...rest } = params;
+  const { isAvailableOnline, filterOutExhibitions, ...rest } = params;
 
   if (params.location)
     locationsValidator({
--- a/api/src/controllers/events.ts
+++ b/api/src/controllers/events.ts
@@ -34,6 +34,7 @@ type QueryParams = {
   location?: string;
   isAvailableOnline?: string;
   timespan?: string;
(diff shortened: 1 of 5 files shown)


rcantin-w (commented):
I would exclude based on `id` and not `label` as these are subject to change and also how we currently do the filtering.

rcantin-w (commented):
I think we need to figure out how often we'll need to exclude things. 
If this is a one-off, I'm thinking we just go for a `filterOutExhibitions=true` type query param. It would never display in the URL as it'd only be used in the `/events*` serverSide queries. I doubt we'd want to filter out other formats.

api/src/controllers/events.ts
```diff
@@ -98,16 +106,29 @@ const paramsValidator = (params: QueryParams): QueryParams => {
   if (params.interpretation)
     prismicIdValidator(params.interpretation, 'interpretations');
   if (params.format) prismicIdValidator(params.format, 'formats');
+  if (params.excludeFormats)
+    prismicIdValidator(params.excludeFormats, 'excludeFormats');
```
rcantin-w, line 110: Are we still supporting this?

rcantin-w (approved):
Aside from the "do we still support `excludeFormats`" question I think it's good to go!

Tested as you suggested and it works well - we can do the final test as part of adding Exhibitions to events, maybe something to add to the Def of Done of that ticket? 

🎉

api/src/controllers/events.ts
gestchild: no, I'll remove it - thanks

PR #249, opened 2025-05-14
Upgrade `express` to v5

rcantin-w:
## What does this change?

This should address the [High security alert](https://github.com/wellcomecollection/content-api/security/dependabot/38) we get for `path-to-regexp`, it is a dependency of `express` and `express@5` uses `path-to-regexp@8`. 

I followed [the migration guide](https://expressjs.com/en/guide/migrating-5.html) and had a look at how we use it - it seems unaffected. I was able to run the API locally.

I ran a fresh `yarn` to make sure a brand new file wouldn't break things and a few TS flags came up which I've addressed here.

## How to test

Run locally too? Remove all modules and lock files and re-run `yarn`; does everything still behave as expected?

No more vulnerabilities

N/A

### DIFF
--- a/pipeline/src/types/prismic/exhibitionHighlightTours.ts
+++ b/pipeline/src/types/prismic/exhibitionHighlightTours.ts
@@ -5,8 +5,6 @@ import {
   PromoSliceZone,
 } from '@weco/content-pipeline/src/types/prismic/';
 
-import { ExhibitionPrismicDocument } from './exhibitions';
-
 type PrismicSlices = {
   primary: {
     title?: { text: string }[];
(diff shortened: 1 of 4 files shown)


pipeline/src/types/prismic/exhibitionHighlightTours.ts
```diff
@@ -23,7 +21,7 @@ type WithExhibitionHighlightTourSlices = {
 export type ExhibitionHighlightTourPrismicDocument = prismic.PrismicDocument<
   {
     related_exhibition: prismic.ContentRelationshipField<
-      ExhibitionPrismicDocument,
+      'exhibitions',
```
rcantin-w, line 24: `ContentRelationshipField` expects a `string` (`TypeEnum - Type API ID of the document.`) and TS complained, so amended it in both places.

PR #277, opened 2025-08-13
Add images and gifVideo to graphQueries

rcantin-w:
## What does this change?

Adds fields to graphquery so we get work links from Editorial Images, Image Galleries and GifVideo

I also added a test to ensure we had linkedWorks representation in our pipeline tests, updating snapshots at the same time.

## How to test

- Article with gifVideo works now has linked works: https://api-dev.wellcomecollection.org/content/v0/all/XP4bGxIAADCGoFOi.articles
- Article with Image gallery now has linked works: https://api-dev.wellcomecollection.org/content/v0/all/ZcDDSBAAACAAKL-_.articles
- Article with Editorial images now has linked works: https://api-dev.wellcomecollection.org/content/v0/all/ZiY2zREAACEAqTaF.articles

All linked works are being fetched

I have encountered some timeouts at reindex time. It wasn't all the time, so I'm at the "weary" stage. As I was adding over a 1000 work fetches, I'm wondering if we wouldn't want to optimise that.

Have I forgotten some fields?

### DIFF
--- a/pipeline/src/transformers/addressables/helpers/catalogue-api.ts
+++ b/pipeline/src/transformers/addressables/helpers/catalogue-api.ts
@@ -1,5 +1,4 @@
 // Helper functions for working with the Wellcome Collection Catalogue API
-import { isNotUndefined } from '@weco/content-pipeline/src/helpers/type-guards';
 
 const getWorkUrl = (workId: string): string => {
   return `https://api.wellcomecollection.org/catalogue/v2/works/${workId}?include=contributors%2Cproduction`;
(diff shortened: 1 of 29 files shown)


pipeline/test/transformers/addressables.test.ts
```diff
@@ -24,6 +24,25 @@ describe('addressables transformer', () => {
     }
   );
 
+  forEachPrismicSnapshot<ArticlePrismicDocument>(['articles'], isAddressable)(
```
gestchild, line 27: We should probably mock the fetchAndTransformWorks function so we aren't making real calls to the catalogue API for the test

gestchild: I can do it in a separate PR

Issue #281, opened 2025-08-19
Addressables are never removed from their index


rcantin-w:
### What

1) `unpublisher` issue: The unpublisher might not (probably doesn't) work as expected for addressables, seeing as their `id` value is modified. So they never get removed from the index and always will show up in search. Currently, [it aims to delete using the `id`](https://github.com/wellcomecollection/content-api/blob/main/unpublisher/src/unpublisher.ts), but the addressables' ids have a different format of `id.[type].[audio|bsl]`

2) `delist` tag: When indexing addressables, [we filter out the ones that are delisted](https://github.com/wellcomecollection/content-api/blob/main/pipeline/src/handler.ts#L23). The thing is, if a document gets tagged as `delist` AFTER it's been indexed, it doesn't get removed or updated as such, since we filter them out. This means they forever show up in search.

### Suggested fix
Add a new `prismicId` field to the addressable document's query object. This field will contain a clean version of the `id`. This will require a mapping change, and therefore, probably a new pipeline (although I think adding a field to the mapping doesn't require a new pipeline, if you want to then edit it for some reason, you won't be able to - up to you).

#### For issue 1: `unpublisher`
Add a check in the `unpublisher` function that checks if it's an addressable query. If so, use the `prismicId` field to target the document(s) to delete (e.g. highlight tours should delete TWO documents, so allow for it). Other queries (articles, events, more to come) should still use the regular `id` as it exists right now.

#### For issue 2: `delist` tag addition
~Create [a new webhook in Prismic](https://prismic.io/docs/webhooks#a-document-is-published) with a "A document is published" trigger. As we want the result to be "delete document in ES if the document has a `delist` tag", we could use the unpublisher function for it too. **Again, this is just for the addressables index.**
If the document has a delist tag, try to delete it from ES (it might not exist) using `prismicId` as above.
If it doesn't, return and do nothing. The document has to stay in the index.~

- include content with the 'delist' tag in the addressables index
- filter out content with the 'delist' tag at query time

Adding or removing the delist tag will be update in the index in the normal way when the document is published in Prismic.

----

Feel free to create sub-issues for these, but they'll both depend on the new `prismicId` field for deletion, so I thought it best to have this under one roof.

PR #296, opened 2025-09-04
Remove addressables from API that have been delisted

gestchild:
## What does this change?

For https://github.com/wellcomecollection/content-api/issues/281

This solves 2 problems with addressables showing up in the content API when they shouldn't.

1. deletes them from the ElasticSearch index when they are unpublished in Prismic (this was previously not working properly)
2. Stops content with a delist tag being returned by the api (content that was delisted after having been published was still showing)

### First part:
- adds a prismicId field to the query objects of all addressables
- added an unpublisher specifically for addressables, which uses deleteByQuery and query.prismicId
- updates types and adds some tests

N.B. I needed to give the elastic_users read permissions for the addressables index in order for deleteByQuery to work.

### Second part:
- indexes all addressables regardless of the presence of the delist tag
- adds the tags to the query objects of all addressables
- filters out addressables when the api queries ElasticSearch

- this PR will also switch the deployment of lambdas and the content api to the new pipeline stuff

### First part
What I did:
- Search for specific addressables in elastic search (in the 2025-09-03 cluster):
`GET addressables/_search
{
"query": {
"term": {
"query.prismicId": "aFAyphIAACAAv-RK"
}
}
}`
- 2 results were returned for exhibition highlight tours (bsl and audio versions)
- [run the unpublisher locally](https://github.com/wellcomecollection/content-api/tree/main/unpublisher#running-unpublisher-locally) with this id (against the 2025-09-03 cluster)
- Run the query again and 0 results returned
- Reindex the addressables to put them back

### Second part
- view all addressables in the 2025-09-03 index
`GET addressables/_search`
- There are 2565 in total
- run the content-api locally
- view all addressables `http://localhost:3002/all`
- There are 2028 in total
- view all addressables with delist tag in the 2025-09-03 index
```
GET addressables/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "term": {
            "query.tags": "delist"
          }
        }
      ]
    }
  }
}
```
- there are 537 in total which makes sense because 2565-2028 = 537, so the content api is filtering out the correct number
- try to retrieve a piece of content with a delist tag using the content-api, http://localhost:3002/all/WXInvioAABlsbLHu.articles
- get a 404
- look for the same piece of content in the 2025-09-03 index `GET addressables/_doc/WXInvioAABlsbLHu.articles`
- get a hit, so we know the content-api is filtering this out

- Unpublished addressable content in Prismic gets removed from the addressables index
- Delisted content in the addressables index is not returned by the content api

It's broken currently so I don't think it can get worse

### DIFF
--- a/unpublisher/README.md
+++ b/unpublisher/README.md
@@ -1,4 +1,4 @@
 ## Running unpublisher locally
 
-Add the pipeline date in a .env file and run:
+Add the pipeline date (`PIPELINE_DATE=20XX-XX-XX`) in a .env file and run:
 `AWS_PROFILE=catalogue-developer yarn dev`


rcantin-w (approved):
I couldn't test if it affected the `/addressables/[id]` query in story pages as they are done client side, so something to quickly check once this is in stage!

Other than my comment about tags versus a boolean, I've nothing to say, looks pretty nifty!

Very much a suggestion bc it's Very Not part of this work would be to add `query.prismicId` to [the fields we list as queryable](https://github.com/wellcomecollection/content-api/blob/main/api/src/queries/addressables.ts). As we list id and uid, could be useful to have the actual prismic id findable as well!

PR #300, opened 2025-09-15
make types more explicit for the addressables

gestchild:
## What does this change?

Relates to https://github.com/wellcomecollection/content-api/pull/299

Makes the addressables types more explicit, we can now only add specific properties to the display property of specific types


rcantin-w (approved):
Nice one! Works well :) (tried to break it and the types are doing a good job now)

PR #310, opened 2025-10-14
Update most packages

rcantin-w:
## What does this change?

Will address all of https://github.com/wellcomecollection/content-api/pull/309, hopefully making the next Dependabot upgrade PR smoother.

All minors/patches were updated without requiring changes. Many major changes were also fine. I'll document below if there was anything special:

Zod changes: https://zod.dev/v4?id=top-level-string-formats
Jest-extended: `toHaveBeenCalledWith` is now supported out of the box so removed now unsupported `toHaveBeenCalledOnceWith`

## How to test

Run the API and run various scripts and tests locally if you want!

Up-to-date on this repo

It's a lot of changes, but we can test it on staging and amend should anything be needed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,8 +14,8 @@ updates:
       - dependency-name: 'eslint*'
       - dependency-name: '@elastic/elasticsearch'
       - dependency-name: 'husky'
-      - dependency-name: '@types/node'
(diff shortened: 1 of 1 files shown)


gestchild (approved):
all seems to be working locally ok

Issue #331, opened 2025-12-10
Replace `filterOutExhibitions`with `format=!exhibitions`


rcantin-w:
(or something like it, but using `!` as an exclusion marker)
This is how the Catalogue API _excludes_, so we should follow that model, it's much more flexible as well. 
This is currently used in wc.org, so both would have to be supported until the change happens on the FE. 
Then, delete `filterOutExhibitions`. 

> [!NOTE]
> Do we have to worry about external people using it?

## Definition of done 
- [ ] Change is done on the FE
- [ ] We don't support `filterOutExhibitions` anymore
- [ ] We can exclude any formats
- [ ] Documentation is updated (developers.wc.org)

PR #338, opened 2025-12-23
Filter exhibitions

gestchild:
## What does this change?

For https://github.com/wellcomecollection/content-api/issues/331

Makes it possible to filter out events by their format type, e.g. format=!{id} or format=!{id-alias}

## How to test

- visit https://api-dev.wellcomecollection.org/content/v0/events and note number of results
- visit https://api-dev.wellcomecollection.org/content/v0/events?filterOutExhibitions=true see fewer results
- visit https://api-dev.wellcomecollection.org/content/v0/events?format=!exhibitions see (the same amount) fewer results

We have a more flexible way of filtering by formats that follows the same pattern as the Catalogue API.

Nothing really


rcantin-w (commented):
https://api-dev.wellcomecollection.org/content/v0/events?filterOutExhibitions is returning 640 results (same as without the filter) but https://api-dev.wellcomecollection.org/content/v0/events?filterOutExhibitions=true works - is that expected?

PR #359, opened 2026-03-03
Upgrade node in prismic-snapshots

rcantin-w:
## What does this change?

Missed this but should be done as part of the whole upgrade.

> [!NOTE]
> This has been applied and we've confirmed it ran properly last night!

I also am updating the `.terraform.lock.hcl` files which are the equivalent of `yarn.lock` for TF.

I'm not sure about what `terraform plan` outputs but to be fair I'd be for applying it and reversing if it implodes. This is for prismic-snapshot, the risk isn't very high.

#### Fixed one test script bug (pre-existing issue)
The local test script for `prismic-backup-download` was already broken on `main` due to a handler signature change in December 2025:
- **Issue**: On Dec 16, 2025, commit [e60c398](https://github.com/wellcomecollection/content-api/commit/e60c398) changed the Lambda handler from expecting `event.batch` to expecting the batch array directly as `event`
- **Fix**: Updated test script to pass `event.batch` to the handler instead of the full event object
- **Note**: This was not caused by the Node upgrade - the test has been failing since December

#### Improved test documentation
- Updated README to use `yarn` commands consistently
- Clarified that tests work without LocalStack (they validate Prismic API integration but fail on S3 operations)
- Documented expected behavior: tests fetch from Prismic successfully but fail on S3 with `InvalidAccessKeyId` errors unless LocalStack is configured
- Commented out LocalStack endpoint by default in test files to match documentation

I'm still not fully convinced by the test coverage, I think there is work to do here, but as a separate concern.

## How to test

Not sure!

Everything on Node 24.

See above


infrastructure/prismic-snapshots/testing/test-prismic-backup-download.js
```diff
@@ -11,10 +11,12 @@ process.env.BUCKET_NAME = 'wellcomecollection-prismic-downloads';
...
   process.env.AWS_SECRET_ACCESS_KEY = 'test';
-  console.log('Using dummy AWS credentials for LocalStack');
+  console.log(
+    'Using dummy AWS credentials (will fail on S3 operations without LocalStack)'
+  );
 }
 
-// For LocalStack, uncomment this:
+// For LocalStack testing with S3, uncomment this:
 process.env.S3_ENDPOINT = 'http://localhost:4566';
```
davidpmccormick: Not introduced in this PR but the comment makes it sounds like this should be commented out?

davidpmccormick (approved):
Just a question about the comment

